PR #45, opened 2021-08-13
Database rework and Startup refactor

LoadingBG:
<details>
<summary>Replace for loops with streams for readability (<code>events/Startup.java</code>)</summary>
For loops can get messy sometimes so I replaced them with streams for a bit more readability in exchange for speed. However, <code>Startup</code> runs only once so the speed doesn't really matter.
</details>
<details>
<summary>Refactor the code a bit (<code>other/Database.java</code>)</summary>
There were some questionable decisions on how to write the code, for example the pattern:
<pre><code>Object o = null;
try {
    o = /* something */;
} catch (Exception e) {
    o = /* default */;
}
return o;
</code></pre> instead of
<pre><code>try {
    return /* something */;
} catch (Exception e) {
    return /* default */;
}
</code></pre>or this inconsistency:
<pre><code>void someMethod() {
    Document setValue = new Document(varName, newValue);
    Document update = new Document("$set", setValue);
    collection.updateOne(query, update);
}
void someOtherMethod() {
    collection.updateOne(query, new BasicDBObject("$set", new BasicDBObject(varName, newValue)));
}
</code></pre>
Also the methods <code>#getConfigString</code>, <code>#getConfigInt</code> and <code>#getConfigBoolean</code> had this weird line:<br/>
<code>if (guildDocExists(guild)) insertGuildDoc(guild);</code><br/>
which I thought was nonsensical because why insert the guild doc if it already exists. I changed it to:<br/>
<code>if (!guildDocExists(guild)) insertGuildDoc(guild);</code>
</details>
<details>
<summary>Make <code>other/Database</code> static</summary>
All the code used random <code>Database</code> instances for one time and then destroyed them. Since <code>Database</code> doesn't hold any internal data, I made all the methods static and made the constructor private.
</details>

### DIFF
--- a/src/main/java/com/javadiscord/javabot/events/Startup.java
+++ b/src/main/java/com/javadiscord/javabot/events/Startup.java
@@ -84,7 +84,7 @@ public class Startup extends ListenerAdapter {
         for (var guild : jdaGuilds) {
             if (skipGuilds.contains(guild.getId())) continue;
 
-            new Database().deleteOpenSubmissions(guild);
+            Database.deleteOpenSubmissions(guild);
             new StarboardListener().updateAllSBM(event, guild);
             Bot.slashCommands.registerSlashCommands(guild);
         }
--- a/src/main/java/com/javadiscord/javabot/events/Startup.java
+++ b/src/main/java/com/javadiscord/javabot/events/Startup.java
@@ -3,20 +3,26 @@ package com.javadiscord.javabot.events;
 import ch.qos.logback.classic.Level;
 import ch.qos.logback.classic.Logger;
 import ch.qos.logback.classic.LoggerContext;
+
 import com.javadiscord.javabot.Bot;
 import com.javadiscord.javabot.commands.other.Version;
 import com.javadiscord.javabot.other.Database;
 import com.javadiscord.javabot.other.Misc;
 import com.mongodb.MongoClient;
 import com.mongodb.MongoClientURI;
+
+import net.dv8tion.jda.api.JDA;
 import net.dv8tion.jda.api.entities.Guild;
 import net.dv8tion.jda.api.events.ReadyEvent;
 import net.dv8tion.jda.api.hooks.ListenerAdapter;
 import org.slf4j.LoggerFactory;
 
+import java.io.BufferedReader;
+import java.io.InputStreamReader;
 import java.util.Arrays;
-import java.util.Scanner;
+import java.util.List;
 import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
 
 
 public class Startup extends ListenerAdapter {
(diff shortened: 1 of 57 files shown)


src/main/java/com/javadiscord/javabot/events/Startup.java
```diff
@@ -40,56 +46,45 @@ public void onReady(ReadyEvent event) {
...
-                sb.append(fReader.nextLine() + "\n");
-            }
-            System.out.println("\n" + sb.toString().replace("{!version}", new Version().getVersion()));
-
+        try (BufferedReader startupMsgReader = new BufferedReader(new InputStreamReader(getClass().getResourceAsStream("textfiles/startup.txt")))) {
+			String startup = startupMsgReader
+                    .lines()
+                    .collect(Collectors.joining("\n"))
+                    .replace("{!version}", new Version().getVersion());
+            System.out.println("\n" + startup);
```
andrewlalis, line 59: If you're going to refactor this, you might as well do `Files.readString(Path)` instead of all that buffered reader stuff.

andrewlalis (changes requested):
I appreciate the effort, but we should not change the database methods to be static. This makes testing individual parts of the application logic very difficult when we can no longer inject a mock of the database somewhere.

If you want to really refactor things, I would appreciate if you could figure out a way to remove the database configuration stuff entirely. There is no point in storing the config in a database when a normal JSON or YAML file will work just as well, especially when the config is read-only for 99.99% of the time.

src/main/java/com/javadiscord/javabot/events/Startup.java
LoadingBG: What is the format of the config?

andrewlalis: It may have changed, but Dynxsty pasted this a while back. The config is all stored in JSON somewhere in a database, but there is not really a reason to store some simple JSON in a database when we can use GSON or Jackson to read from a file.

Also, I would suggest maybe being a bit patient with this change, because I've also been pushing him to upgrade to java 16 which adds records, which would be really handy for config.
```json
{
    "name": "Java",
    "guild_id": "648956210850299986",
    "welcome_system": {
        "image": {
            "avatar": {
                "avX": 75,
                "avY": 100,
                "avH": 400,
                "avW": 400
            },
            "imgW": 1920,
            "imgH": 600,
            "overlayURL": "https://cdn.discordapp.com/attachments/744899463591624815/827303132098461696/WelcomeOverlay_NoShadow.png",
            "bgURL": "https://cdn.discordapp.com/attachments/744899463591624815/840928322661122048/WelcomeBG.png",
            "primCol": "16777215",
            "secCol": "16720173"
        },
        "join_msg": "Hey {!member}, welcome to **{!server}**!",
        "leave_msg": "**{!membertag}**{!boticon} left the server.",
        "welcome_cid": "650632933165760512",
        "welcome_status": true
    },
    "channels": {
        "report_cid": "760172343971807244",
        "log_cid": "793486198584115270",
        "suggestion_cid": "752535909228085348",
        "submission_cid": "811277208223875112",
        "jam_announcement_cid": "796057592405032990",
        "jam_vote_cid": "796057702635274270"
    },
    "roles": {
        "mute_rid": "736591398849085460",
        "jam_ping_rid": "761320255082463263",
        "jam_admin_rid": "735510411377508411"
    },
    "other": {
        "stats_category": {
            "stats_cid": "773622646876274728",
            "stats_text": "Server · {!membercount} Members"
        },
        "qotw": {
            "dm-qotw": true
        },
        "server_lock": {
            "lock_status": false,
            "lock_count": 0
        },
        "starboard": {
            "starboard_cid": "866099127821402112",
            "starboard_emote": "⭐",
            "starboard_emote2": "🌟",
            "starboard_emote3": "🌠"
        }
    }
}
```

PR #54, opened 2021-08-18
Refactor other/Database (try 2)

LoadingBG:
<details>
<summary>Add Eclipse files to gitignore</summary>
I use Eclipse LSP for autocompletion and I'm tired of trying to not commit the files so I added them to the gitignore
</details>
<details>
<summary>Refactor <code>other/Database.java</code></summary>
This time I just refactored the class, without making a bunch of static methods. Also I removed the usage of GSON because it's not needed.
</details>
<details>
<summary>Refactor <code>other/StarboardListener</code></summary>
<ul>
<li>Removed usage of GSON</li>
<li>Most database-related methods were moved to <code>other/Database</code></li>
<li>
Some changes to methods:
<ul>
<li><code>#setEmoteCount(int, String, String, String)</code> -> <code>#setEmoteCount(String, String, String, int)</code>: to match other methods which accept guild ID, channel ID and message ID first</li>
<li><code>StarboardListener#updateAllSBM(ReadyEvent, Guild)</code> -> <code>StarboardListener#updateAllSBM(Guild)</code>: the <code>ReadyEvent</code> was used like this: <code>event.getJDA().getGuildById(guild.getId())</code> which is useless because it just gets the <code>Guild</code> instance already passed in.</li>
</ul>
</li>
</ul>
</details>

### DIFF
--- a/src/main/java/com/javadiscord/javabot/events/StarboardListener.java
+++ b/src/main/java/com/javadiscord/javabot/events/StarboardListener.java
@@ -30,7 +30,7 @@ public class StarboardListener extends ListenerAdapter {
         String messageId = channel.getId();
 
         Database db = new Database();
-        db.changeSBCBool(guildId, channelId, messageId, true);
+        db.changeStarboardChannelBool(guildId, channelId, messageId, true);
         TextChannel sc = guild.getTextChannelById(db.getConfigString(guild, "other.starboard.starboard_cid"));
 
         EmbedBuilder eb = new EmbedBuilder()
@@ -51,7 +51,7 @@ public class StarboardListener extends ListenerAdapter {
                 e.printStackTrace();
             }
         }
-        msgAction.queue(sbMsg -> db.querySBDString(guildId, channelId, messageId, "starboard_embed", sbMsg.getId()));
+        msgAction.queue(sbMsg -> db.queryStarboardString(guildId, channelId, messageId, "starboard_embed", sbMsg.getId()));
     }
 
     void updateSB(Guild guild, String cID, String mID) {
@@ -59,7 +59,7 @@ public class StarboardListener extends ListenerAdapter {
         Database db = new Database();
         String gID = guild.getId();
 
-        String sbcEmbedId = db.getSBCString(gID, cID, mID, "starboard_embed");
+        String sbcEmbedId = db.getStarboardChannelString(gID, cID, mID, "starboard_embed");
         if (sbcEmbedId == null) {
             return;
         }
@@ -82,7 +82,7 @@ public class StarboardListener extends ListenerAdapter {
                     + starCount + " | " + tc.getAsMention()).queue();
         } else {
             sbMsg.delete().queue();
-            db.deleteSBMessage(gID, cID, mID);
+            db.deleteStarboardMessage(gID, cID, mID);
         }
     }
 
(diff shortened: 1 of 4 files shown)


src/main/java/com/javadiscord/javabot/events/StarboardListener.java
```diff
@@ -27,326 +25,201 @@
...
+        String messageId = channel.getId();
 
-        changeSBCBool(guild.getId(), channel.getId(), message.getId(), true);
-        TextChannel sc = guild.getTextChannelById(new Database().getConfigString(guild, "other.starboard.starboard_cid"));
+        Database db = new Database();
+        db.changeSBCBool(guildId, channelId, messageId, true);
+        TextChannel sc = guild.getTextChannelById(db.getConfigString(guild, "other.starboard.starboard_cid"));
 
-        var eb = new EmbedBuilder()
+        EmbedBuilder eb = new EmbedBuilder()
```
andrewlalis, line 36: Changing `var` to another variable type is something that doesn't really need to be changed. Either form is acceptable, so leave these as-is, otherwise it might turn into an edit war where people change it back and forth each time.

src/main/java/com/javadiscord/javabot/other/Database.java
```diff
@@ -24,65 +23,47 @@
 
     private static final org.slf4j.Logger logger = LoggerFactory.getLogger(Database.class);
 
-    private static final String[] databases = new String[]{"userdata", "other"};
-    private static final String[] userdataCollections = new String[]{"potential_bot_list", "users", "warns"};
-    private static final String[] otherCollections = new String[]{"config", "customcommands", "expert_questions", "open_submissions", "reactionroles", "starboard_messages", "submission_messages"};
+    private static final Map<String, String[]> DB_COLS = new HashMap<>();
```
andrewlalis, line 26: It would be nice to have either
- a more descriptive name of what this variable is for, or...
- javadoc description of this variable.

src/main/java/com/javadiscord/javabot/other/Database.java
```diff
@@ -169,253 +145,245 @@ public Document guildDoc (Guild guild) {
...
+        Document query = new Document("guild_id", guildId)
+                .append("channel_id", channelId)
+                .append("message_id", messageId);
+        mongoClient.getDatabase("other")
+                .getCollection("starboard_messages")
+                .updateOne(query, update);
     }
 
-    public String getConfigChannelAsMention (Guild guild, String path) {
+    public void querySBDString(String gID, String cID, String mID, String value, String newValue) {
```
andrewlalis, line 303: For this and the other `...SB` methods, it would be nice to refactor their names to be `...StarBoard` instead, so that it's actually readable for everyone.

andrewlalis (changes requested):
Just a few small comments regarding things that should be improved before merging. However, some overall feedback:

I think that we're wasting effort by refactoring the database stuff, when this sort of thing should not be stored in a database in the first place. It would be much more efficient, and easier for everyone involved if we moved to a file-based config. That way, we would not require developers to have their own instance of MongoDB just to run the most core bot functionality. Also, now that we're in Java 16, we can easily use record DTOs for config, which removes the antipattern of having to use string literals to fetch data all over the place.

src/main/java/com/javadiscord/javabot/other/Database.java
LoadingBG: Almost all methods need a javadoc.

src/main/java/com/javadiscord/javabot/other/Database.java
LoadingBG: sure

andrewlalis (approved):

PR #59, opened 2021-08-23
Make `properties/command` classes records

LoadingBG:
Make `CommandConfig`, `CommandPrivilegeConfig`, `OptionConfig`, `SubCommandConfig` and `SubCommandGroupConfig` records since `lombok.Data` annotation almost the same thing and the config doesn't need to be mutated.
- All fields of an array type were turned into instances of `List` because `record` doesn't support arrays as well in `#equals` and `#hashCode`


andrewlalis:
@LoadingBG have you tested this properly? I am vaguely aware of Jackson/SnakeYAML having some issues with deserialization of generic types, since at runtime it doesn't now what objects the List contains.

PR #62, opened 2021-08-28
main

sahilasopa:


andrewlalis:
We're currently not looking for improvements to the configuration options; I'm working on a way to do this without a database dependency.

PR #77, opened 2021-09-26
Add Automated Help Channel Moderation System [Beta]

### LINKED ISSUE #1: Bot-moderated help channel automation
Consider adding bot-managed help channels to streamline people's questions and answers.

### Proposed Flow

There are several open help channels that are clearly visible. Users are directed to ask their question in one of these channels.

Once the user sends their first message in an open channel, the channel is moved to a separate area for "active questions". It will stay open in a "pending answer" state until another user (not the original question asker) sends a message in the channel, as a response to the question. If no response is given after X hours, the bot sends a message in the channel to ask the user to confirm that the question is still unanswered, and the channel will stay reserved. Repeat this confirmation flow as long as possible (or introduce a hard cutoff limit).

Once some discussion has happened in a reserved help channel, and the user has obtained their answer, the bot will wait for at least X minutes of inactivity, after which it will send a question to the user in the channel, asking if their question was successfully resolved. If they respond affirmatively, the channel is moved back to the list of available channels, or if not, it stays "reserved".

### Channel States
- **Available** - The channel is available for users to ask questions.
- **Pending Answer** - The channel has been reserved by a user and they've asked their question, and it is waiting for a response from another user.
- **Active** - Once another user responds, channels are active while any user sends a message within X minutes.
- **Inactive** - If there's been no activity in an active channel in X minutes, the channel is considered inactive. After some time, the bot will confirm with the original question asker if they got their question answered. If so, channel returns to **Available** state. After a long time, the channel will return to **Available** regardless.

andrewlalis:
Closes #1

### DIFF
--- a/src/main/java/com/javadiscord/javabot/properties/config/guild/HelpConfig.java
+++ b/src/main/java/com/javadiscord/javabot/properties/config/guild/HelpConfig.java
@@ -1,12 +1,34 @@
 package com.javadiscord.javabot.properties.config.guild;
 
+import com.javadiscord.javabot.help.AlphabetNamingStrategy;
+import com.javadiscord.javabot.help.AnimalNamingStrategy;
+import com.javadiscord.javabot.help.ChannelNamingStrategy;
 import com.javadiscord.javabot.properties.config.GuildConfigItem;
 import lombok.Data;
 import lombok.EqualsAndHashCode;
+import net.dv8tion.jda.api.entities.Category;
 
+/**
+ * Configuration for the guild's help system.
+ */
 @Data
 @EqualsAndHashCode(callSuper = true)
 public class HelpConfig extends GuildConfigItem {
+	/**
+	 * The id of the channel category that all help channels are in.
+	 */
+	private long categoryId;
+
+	/**
+	 * The strategy to use when naming help channels.
+	 */
+	private String channelNamingStrategy = "animal";
+
+	/**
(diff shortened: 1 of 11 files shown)


src/main/java/com/javadiscord/javabot/properties/config/guild/HelpConfig.java
```diff
@@ -0,0 +1,79 @@
...
+	 * The string which is prefixed to any open help channel, where users are
+	 * free to ask a question.
+	 */
+	private String openChannelPrefix = "\uD83D\uDFE2";
+
+	/**
+	 * The string which is prefixed to any reserved help channel, where a user
+	 * has already asked a question and is in the process of getting an answer.
+	 */
+	private String reservedChannelPrefix = "⛔";
```
danthe1st, line 42: openChannelPrefix uses escaped unicode sequences and reservedChannelPrefix uses a unicode emoji directly embedded in the code.

danthe1st (changes requested):
I don't think embedded and escaped unicode emojis should be mixed.

src/main/java/com/javadiscord/javabot/properties/config/guild/HelpConfig.java
andrewlalis: It is fixed.

danthe1st (approved):

PR #88, opened 2021-10-03
Moved Stats Category updates to a scheduled Thread pool

DynxstyGIT:


src/main/java/com/javadiscord/javabot/StatsUpdater.java
```diff
@@ -0,0 +1,44 @@
...
+import net.dv8tion.jda.api.events.ReadyEvent;
+import net.dv8tion.jda.api.hooks.ListenerAdapter;
+
+import javax.annotation.Nonnull;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+
+@Slf4j
+public class StatsUpdater extends ListenerAdapter {
```
andrewlalis, line 15: Can you add some javadoc here explaining what this class does?

src/main/java/com/javadiscord/javabot/StatsUpdater.java
```diff
@@ -0,0 +1,44 @@
...
+
+    private final ScheduledExecutorService threadPool = Executors.newSingleThreadScheduledExecutor();
```
andrewlalis, line 17: You should use `Bot.asyncPool` instead of creating another thread pool here. That thread pool already has I think 4 threads so it should be more than enough.

src/main/java/com/javadiscord/javabot/StatsUpdater.java
```diff
@@ -0,0 +1,44 @@
...
+
+    private final long delay = 20;
```
andrewlalis, line 19: This should probably be moved to a config setting in `StatsConfig`

src/main/java/com/javadiscord/javabot/StatsUpdater.java
```diff
@@ -0,0 +1,44 @@
...
+
+    private final TimeUnit delayUnit = TimeUnit.MINUTES;
+
+    private JDA jda;
```
andrewlalis, line 23: I don't think there's any reason to keep a reference to JDA here, since you only ever use it in the onReady event where it's originally obtained.

src/main/java/com/javadiscord/javabot/StatsUpdater.java
```diff
@@ -0,0 +1,44 @@
...
+import java.util.concurrent.TimeUnit;
+
+@Slf4j
+public class StatsUpdater extends ListenerAdapter {
+
+    private final ScheduledExecutorService threadPool = Executors.newSingleThreadScheduledExecutor();
+
+    private final long delay = 20;
+
+    private final TimeUnit delayUnit = TimeUnit.MINUTES;
```
andrewlalis, line 21: I don't really think this needs to be defined as a class constant if it's only used in one place.

andrewlalis (changes requested):

src/main/java/com/javadiscord/javabot/StatsUpdater.java
DynxstyGIT: But wouldnt that mean that we need a separate threadpool for each guild?

andrewlalis (approved):

PR #95, opened 2021-10-07
Share-knowledge voting System.

cbrt-x:


Jadefalke2:
#95 
Only the channel ID and the treshold need to be adjusted

PR #98, opened 2021-10-08
Added Stackoverflow command

sahilasopa:


Kemikals:
> > As a possible alternative, the [StackExchange Search API](https://api.stackexchange.com/docs/search) can be used.
> > Example: https://api.stackexchange.com/search?site=stackoverflow&intitle=hello%20world
> 
> i actually saw the api but the issue with that is it wont match exactly, for example i entered "how to take input in java" my results were links to questions like "How to take the this input in 5x5 char matrix in java ?" so beginners of this server might find it less helpful in othercase if you just do a google search the most relevant answer pops up

Did you take a peek at the advanced version of the API? changing a few options like sort by relevance, discarding closed, and adding the java tag makes it pretty accurate I think can play with it here

https://api.stackexchange.com/docs/advanced-search#order=desc&sort=relevance&q=how%20to%20take%20input%20in%20java&accepted=True&closed=False&tagged=java&filter=default&site=stackoverflow&run=true

DynxstyGIT:
@sahilasopa

MoonTM-GIT:
Yeah so I agree with what dan1st said that we don't need a Google & lmgtfy command, if we decide to add that Google command we should remove the other one

MoonTM-GIT:
bruh I fucked things up

MoonTM-GIT:
didnt even do anything on this pr and it somehow merged it and I reverted something else I did and now everything's broken 😤

Issue #99, opened 2021-10-09
Allow Helpers to unreserve help channels


MoonTM-GIT:
https://canary.discord.com/channels/648956210850299986/752535909228085348/896417630607523910

PR #102, opened 2021-10-10
Helpers unreserve channels

Kemikals:
#99 removed role config


Kemikals:
> Looks good, only thing is that you should maintain the convention of tab indentation instead of spaces.

Fixed, should we make a ticket for standardizing the style across the project? 

> Looks good, only thing is that you should maintain the convention of tab indentation instead of spaces.

Fixed. Should there be an issue for maybe creating a contributing.md that lists the style for the project. And maybe a task to bring the project up to the proposed style? There's quite a few places in the project using 4 spaces over tabs

danthe1st:
Maybe also create a `. editorconfig` file.

PR #129, opened 2021-12-05
fix possibility to thank same user multiple times

danthe1st:
Currently, it is possible to thank the same user multiple times in the same question.
In order to do this, one needs to use the `/unreserve` command and thank the user.
After that, they can use `/unreserve` once again bringing up the menu to thank the user a second time.
This can be repeated over and over.

This PR fixes this issue by deleting the unreservation message (with the thanks-buttons) and disabling the new buttons for users already thanked.

### DIFF
--- a/src/main/java/com/javadiscord/javabot/service/help/HelpChannelManager.java
+++ b/src/main/java/com/javadiscord/javabot/service/help/HelpChannelManager.java
@@ -9,6 +9,7 @@ import com.javadiscord.javabot.utils.MessageActionUtils;
 import lombok.extern.slf4j.Slf4j;
 import net.dv8tion.jda.api.entities.*;
 import net.dv8tion.jda.api.interactions.Interaction;
+import net.dv8tion.jda.api.interactions.components.Button;
 import net.dv8tion.jda.api.interactions.components.ButtonStyle;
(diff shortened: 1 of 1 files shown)


andrewlalis:
How recently did you test this? I added a check that should only insert a new record (thank entity) if none yet exists for the unique combination of help channel reservation id and helper id.

danthe1st:
I tested it with the state of yesterday I think

andrewlalis:
> I tested it with the state of yesterday, merged it (fixing some conflicts) and tested it again (but just normal unreserve and multiple unreservations with thanking users). But why would you use reservation IDs anyways since my approach works without saving any extra data.

Because having unique ids for each help session is the cleanest way to 100% guarantee that there can never be any instance of someone being thanked twice; we're defining the data model to forbid it from ever happening. So it's a cleaner approach in the long term.

danthe1st:
Is that already fully implemented?

andrewlalis:
> Is that already fully implemented?

Yeah, this was the main commit where I did that, followed by a few smaller fix/improvement commits: https://github.com/Java-Discord/JavaBot/commit/1a0b59d636bab1e81e5202e9eee01c6b28f5dd51

danthe1st:
already implemented

PR #132, opened 2021-12-15
Added check for known scam links to Automod

### LINKED ISSUE #86: Improve Auto mod to detect phishing links
e.g. _and probably most important for now_ links looking like discord.com that aren't real

Could also include a general list of known phishing links

MoonTM-GIT:
Added support for known scam links to Automod. List of links from https://github.com/DevSpen/links

Closes #86


src/main/java/com/javadiscord/javabot/Bot.java
```diff
@@ -55,6 +59,11 @@
      */
     public static HikariDataSource dataSource;
 
+    /**
+     * A list of known links used for scamming. (from https://github.com/DevSpen/links)
+     */
+    public static List<String> suspiciousLinks;
```
andrewlalis, line 65: I don't think that this should be done here, in the bot's main class, because these suspicious links aren't something that will be used by tons of different components. It would be better if you just loaded them in the AutoMod's constructor or something.

andrewlalis (changes requested):

MoonTM-GIT:
Yeah I already know about my mistake, you don't need to look at it rn

src/main/java/com/javadiscord/javabot/Bot.java
MoonTM-GIT: Yeah, I'll fix that later today. Dynxsty also said, that it won't work because if compiled to a jar there's no src folder so I need to fix that as well

MoonTM-GIT:
As @sahilasopa is working on his implementation again we can probably close this

PR #137, opened 2021-12-16
Added `/search` command and automatic scam link detection 

sahilasopa:


src/main/java/com/javadiscord/javabot/service/AutoMod.java
```diff
@@ -85,6 +108,15 @@ private void checkContentAutomod(@Nonnull Message message) {
         if (matcher.find()) {
             warn(message, message.getMember(), "Automod: Advertising");
         }
+        final String messageRaw = message.getContentRaw();
+        if (messageRaw.startsWith("http://") || messageRaw.startsWith("https://")) {
+            // only do it for a links, so it won't iterate for each message
+            for (String spamUrl : spamUrls) {
+                if (messageRaw.contains(spamUrl)){
+                    new Ban().ban(message.getMember(), "Scam");
```
MoonTM-GIT, line 116: You're still banning people although we said we would warn them

And please use the same reason-naming-scheme we used everywhere else for AutoMod, so it should be `Automod: Suspicious Link` or something similar

MoonTM-GIT: People are stupid and they will post links on accident or because they want to know whether it's harmful or not and just don't know the damages it might cause to others - we don't ban people if they post a message with that link asking whether it's safe or not - we tell them not to do it again and in some cases warn them.

MoonTM-GIT: And something else regarding this part; You should also delete the message that was sent

sahilasopa: Alright Done

MoonTM-GIT (approved):

danthe1st:
Are we sure we need a `/google` **and** a `/lmgtfy` command?

sahilasopa:
> Are we sure we need a `/google` **and** a `/lmgtfy` command?

As suggested by @MoonTM-GIT we will be removing/lmgfy and replacing it with 
/google

DynxstyGIT:
Didn't @MoonTM-GIT already implement some kind of link detection?

MoonTM-GIT:
Yes, but them Sahil said that he had already made one in his old PR so I closed my PR and we decided to use this one

src/main/java/com/javadiscord/javabot/service/AutoMod.java
```diff
@@ -85,6 +108,20 @@ private void checkContentAutomod(@Nonnull Message message) {
         if (matcher.find()) {
             warn(message, message.getMember(), "Automod: Advertising");
         }
+        final String messageRaw = message.getContentRaw();
+        if (messageRaw.startsWith("http://") || messageRaw.startsWith("https://")) {
+            // only do it for a links, so it won't iterate for each message
+            for (String spamUrl : spamUrls) {
+                if (messageRaw.contains(spamUrl)){
+                    try {
+                        message.delete();
```
DynxstyGIT, line 117: This need to be queued.

src/main/resources/commands.yaml
```diff
@@ -30,6 +30,12 @@
     - {description: Text that will be converted into a lmgtfy-link, name: text, required: true,
        type: STRING}
   handler: com.javadiscord.javabot.commands.user_commands.Lmgtfy
+- description: Generates A Google Link For Given Input
+  name: google
+  options:
+    - {description: Text that will be converted into a google-link, name: text, required: true,
+       type: STRING}
```
DynxstyGIT, line 37: Uncollapse this

src/main/java/com/javadiscord/javabot/commands/user_commands/Google.java
```diff
@@ -0,0 +1,27 @@
...
+        String query;
+        String url;
+        try {
+            query = URLEncoder.encode(Objects.requireNonNull(event.getOption("query")).getAsString(), StandardCharsets.UTF_8.toString());
+        } catch (UnsupportedEncodingException e) {
+            e.printStackTrace();
+            return event.reply(e.getMessage());
+        }
+        query = query.trim().replace(" ", "+");
+        url = "https://www.google.com/search?q=".concat(query);
```
DynxstyGIT, line 24: Instead of just returning the raw search url, wouldn't it be better to list the top 3/5 search results instead? 
Guess google provides some sort of API for this.

DynxstyGIT (changes requested):

src/main/java/com/javadiscord/javabot/commands/user_commands/Google.java
```diff
@@ -0,0 +1,27 @@
...
+public class Google implements SlashCommandHandler {
+    @Override
+    public ReplyAction handle(SlashCommandEvent event) {
+        String query;
+        String url;
+        try {
+            query = URLEncoder.encode(Objects.requireNonNull(event.getOption("query")).getAsString(), StandardCharsets.UTF_8.toString());
+        } catch (UnsupportedEncodingException e) {
+            e.printStackTrace();
+            return event.reply(e.getMessage());
```
MoonTM-GIT, line 21: Please use the [Responses](https://github.com/Java-Discord/JavaBot/blob/main/src/main/java/com/javadiscord/javabot/commands/Responses.java) class for this

src/main/java/com/javadiscord/javabot/service/AutoMod.java
```diff
@@ -24,6 +31,22 @@
...
+            Scanner scanner = new Scanner(new File(String.valueOf(getClass().getResourceAsStream("spamsLinks.txt"))));
+            spamUrls = new ArrayList<>();
+            while (scanner.hasNext()){
+                spamUrls.add(scanner.next());
+            }
+            scanner.close();
+        } catch (FileNotFoundException e) {
+            e.printStackTrace();
+        }
+
```
DynxstyGIT, line 47: I guess a better/cleaner approach would be something like this:

```java
try (var linesStream = Files.lines(Paths.get(getClass().getResource("/spamLinks.txt").toURI()))) {
            linesStream.forEach(spamUrls::add);
        } catch (IOException | URISyntaxException e) {
            e.printStackTrace();
        }
```

DynxstyGIT (changes requested):

src/main/java/com/javadiscord/javabot/service/AutoMod.java
```diff
@@ -24,6 +31,22 @@
  */
 public class AutoMod extends ListenerAdapter {
 
+    private static List<String> spamUrls;
+
+    public AutoMod() {
+        try {
+            Scanner scanner = new Scanner(new File(String.valueOf(getClass().getResourceAsStream("spamsLinks.txt"))));
```
MoonTM-GIT, line 38: And on your last PR Andrew said you shouldn't use a Scanner for this; As he didn't want me to change my approach you _could_ use [Files#getAllLines](https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#readAllLines-java.nio.file.Path-)

MoonTM-GIT (changes requested):

MoonTM-GIT:
And could you please remove the lmgtfy command in this PR as well?

src/main/java/com/javadiscord/javabot/commands/user_commands/Google.java
DynxstyGIT: https://serpapi.com/ ?

src/main/resources/commands.yaml
sahilasopa: What do you mean by `uncollapse this`?

DynxstyGIT: do it like that:
```yaml
- name: text
  description: Text that will be converted into a google-link
  required: true
  type: STRING
```

src/main/java/com/javadiscord/javabot/service/AutoMod.java
```diff
@@ -34,14 +33,10 @@
...
-        try {
-            Scanner scanner = new Scanner(new File(String.valueOf(getClass().getResourceAsStream("spamsLinks.txt"))));
-            spamUrls = new ArrayList<>();
-            while (scanner.hasNext()){
-                spamUrls.add(scanner.next());
-            }
-            scanner.close();
-        } catch (FileNotFoundException e) {
+        spamUrls = new ArrayList<>();
+        try (var linesStream = Files.lines(Paths.get(getClass().getResource("/spamLinks.txt").toURI()))) {
```
MoonTM-GIT, line 37: Just a suggestion because I think you didn't see what I had said; this is what Dynxsty suggested but I'm of the opinion that what I said would be the better approach.

src/main/java/com/javadiscord/javabot/commands/user_commands/Google.java
MoonTM-GIT: As Google doesn't provide an API for stuff like this we have decided to rename the command to `/search` and use the [Bing Search API](https://www.microsoft.com/en-us/bing/apis/bing-web-search-api)

src/main/java/com/javadiscord/javabot/service/AutoMod.java
DynxstyGIT: yeah

MoonTM-GIT (approved):

src/main/java/net/javadiscord/javabot/systems/commands/SearchCommand.java
```diff
@@ -0,0 +1,95 @@
...
+import net.javadiscord.javabot.command.Responses;
+import net.javadiscord.javabot.command.SlashCommandHandler;
+import net.javadiscord.javabot.data.config.BotConfig;
+
+public class SearchCommand implements SlashCommandHandler {
+    static BotConfig config = new BotConfig(Path.of("config"));
+    static String subscriptionKey = config.getSystems().azureSubscriptionKey;
+
+    static String host = "https://api.bing.microsoft.com";
+    static String path = "/v7.0/search";
```
DynxstyGIT, line 24: This stuff doesn't need to be static. Convert the `subscriptionKey` to a local variable and make the other stuff private and final. And since the host- & path strings are constants, change their name to `HOST` & `PATH` to match Javas naming conventions

src/main/java/net/javadiscord/javabot/systems/commands/SearchCommand.java
```diff
@@ -0,0 +1,95 @@
...
+            StringBuilder resultString = new StringBuilder();
+            for (int i = 0; i < 3; i++) {
+                JsonObject object = urls.get(i).getAsJsonObject();
+                name = object.get("name").getAsString();
+                url = object.get("url").getAsString();
+                snippet = object.get("snippet").getAsString();
+                if (object.get("snippet").getAsString().length() > 260) {
+                    snippet = object.get("snippet").getAsString().substring(0, 260).concat("...");
+                }
+                resultString.append("**" + (i+1) + ". [" + name  + "](" + url +")** \n" + snippet + "\n\n");
```
DynxstyGIT, line 76: Use multiple `.append` methods there instead

src/main/java/net/javadiscord/javabot/systems/commands/SearchCommand.java
```diff
@@ -0,0 +1,95 @@
...
+
+    static class SearchResults {
+        HashMap<String, String> relevantHeaders;
+        String jsonResponse;
+
+        SearchResults(HashMap<String, String> headers, String json) {
+            relevantHeaders = headers;
+            jsonResponse = json;
+        }
+    }
```
DynxstyGIT, line 94: You could use a record instead

src/main/java/net/javadiscord/javabot/systems/moderation/AutoMod.java
```diff
@@ -99,6 +117,28 @@ private void checkContentAutomod(@Nonnull Message message) {
                             false
                     );
         }
+        final String messageRaw = message.getContentRaw();
+        if (messageRaw.startsWith("http://") || messageRaw.startsWith("https://")) {
```
DynxstyGIT, line 121: maybe replace `.startsWith()` with `.contains()`

src/main/java/net/javadiscord/javabot/systems/moderation/AutoMod.java
```diff
@@ -22,6 +28,18 @@
  */
 public class AutoMod extends ListenerAdapter {
 
+    private static List<String> spamUrls;
+
+    public AutoMod() {
+        spamUrls = new ArrayList<>();
```
DynxstyGIT, line 34: I think you can remove that line entirely

src/main/java/net/javadiscord/javabot/systems/moderation/AutoMod.java
```diff
@@ -22,6 +28,18 @@
  */
 public class AutoMod extends ListenerAdapter {
 
+    private static List<String> spamUrls;
```
DynxstyGIT, line 31: Again, this doesn't need to be static

src/main/java/net/javadiscord/javabot/systems/commands/SearchCommand.java
```diff
@@ -0,0 +1,95 @@
...
+                .setColor(Bot.config.get(event.getGuild()).getSlashCommand().getDefaultColor())
+                .setTitle("Search Results");
+
+        try {
+            SearchResults result = SearchWeb(searchTerm);
+            JsonObject json = JsonParser.parseString(result.jsonResponse).getAsJsonObject();
+            JsonArray urls = json.get("webPages").getAsJsonObject().get("value").getAsJsonArray();
+            StringBuilder resultString = new StringBuilder();
+            for (int i = 0; i < 3; i++) {
+                JsonObject object = urls.get(i).getAsJsonObject();
```
DynxstyGIT, line 69: check if there are at least 3 search results. You could use the `Math.min(int, int)` method for this:
```java
for (int i = 0; i < Math.min(3, urls.size); i++) {
...
}
```

src/main/java/net/javadiscord/javabot/systems/commands/SearchCommand.java
```diff
@@ -0,0 +1,95 @@
...
+                results.relevantHeaders.put(header, headers.get(header).get(0));
+            }
+        }
+        stream.close();
+        return results;
+    }
+
+    @Override
+    public ReplyAction handle(SlashCommandEvent event) {
+        String searchTerm = event.getOption("query").getAsString();
```
DynxstyGIT, line 55: Add a null-check here, like here:
https://github.com/Java-Discord/JavaBot/blob/63ab4f11d7d68f9e304c8ea91d3a417f9142a639/src/main/java/net/javadiscord/javabot/systems/commands/RegexCommand.java#L19-L26

DynxstyGIT (changes requested):

src/main/java/net/javadiscord/javabot/systems/commands/SearchCommand.java
```diff
@@ -17,19 +17,17 @@
 import net.javadiscord.javabot.data.config.BotConfig;
 
 public class SearchCommand implements SlashCommandHandler {
-    static BotConfig config = new BotConfig(Path.of("config"));
-    static String subscriptionKey = config.getSystems().azureSubscriptionKey;
-
-    static String host = "https://api.bing.microsoft.com";
-    static String path = "/v7.0/search";
+    private final BotConfig config = new BotConfig(Path.of("config"));
```
DynxstyGIT, line 20: You could use the static `BotConfig` variable from the Bot class instead

PR #149, opened 2021-12-27
Fix `checkstyle` problems

MoonTM-GIT:
My IDE also added an empty line between the description and `@param`, `@returns`, etc. tags but I think this might actually be a good thing, especially when you look at bigger texts. I hope I didn't miss any other things it changed, but yeah maybe someone else can spot something


DynxstyGIT:
Looks good, could we use that PR to fix all checkstyle problems tho?

DynxstyGIT (approved):

MoonTM-GIT:
sure

PR #153, opened 2021-12-29
Added more suspicious links

sahilasopa:
added more phishing links in spamLinks.txt


MoonTM-GIT:
Where did you find these?

sahilasopa:
> Where did you find these?

https://github.com/DevSpen/scam-links/blob/master/src/links.txt

danthe1st:
The bot could fetch the list of links on startup (and caching them if fetching doesn't work) instead of relying on PRs whenever the list is changed.

sahilasopa:
> The bot could fetch the list of links on startup (and caching them if fetching doesn't work) instead of relying on PRs whenever the list is changed.

Yup, I was thinking the same

PR #159, opened 2022-01-05
Fixed bug causing unnecessary warns

sahilasopa:
Added regex to extract the link from the url and validate if its a exact match
the earlier bug was caused by
`https://steamcommunity.com` containing the url `https://steamcommunity.co` which is an suspicious url


src/main/java/net/javadiscord/javabot/systems/moderation/AutoMod.java
```diff
@@ -30,6 +30,11 @@
 public class AutoMod extends ListenerAdapter {
 
 	private static final Pattern inviteURL = Pattern.compile("discord(?:(\\.(?:me|io|gg)|sites\\.com)/.{0,4}|app\\.com.{1,4}(?:invite|oauth2).{0,5}/)\\w+");
+	Pattern urlPattern = Pattern.compile(
```
danthe1st, line 33: Why is this package private?

danthe1st: I meant you may want to make it private, not make other variables package private, too.

src/main/java/net/javadiscord/javabot/systems/moderation/AutoMod.java
```diff
@@ -29,15 +34,29 @@
 @Slf4j
 public class AutoMod extends ListenerAdapter {
 
-	private static final Pattern inviteURL = Pattern.compile("discord(?:(\\.(?:me|io|gg)|sites\\.com)/.{0,4}|app\\.com.{1,4}(?:invite|oauth2).{0,5}/)\\w+");
+	final Pattern inviteURL = Pattern.compile("discord(?:(\\.(?:me|io|gg)|sites\\.com)/.{0,4}|app\\.com.{1,4}(?:invite|oauth2).{0,5}/)\\w+");
+	Pattern urlPattern = Pattern.compile(
+			"(?:^|[\\W])((ht|f)tp(s?)://|www\\.)"
+					+ "(([\\w\\-]+\\.)+?([\\w\\-.~]+/?)*"
```
DynxstyGIT, line 40: Please make that Pattern private

src/main/java/net/javadiscord/javabot/systems/moderation/AutoMod.java
```diff
@@ -29,15 +34,29 @@
...
+			spamUrls = List.of(response.split("\n"));
+		} catch (IOException e) {
 			e.printStackTrace();
-			spamUrls = List.of();
+			try {
+				spamUrls = Arrays.stream(StringResourceCache.load("/spamLinks.txt").split(System.lineSeparator())).toList();
+			} catch (Exception exception) {
+				exception.printStackTrace();
+				spamUrls = List.of();
+			}
```
DynxstyGIT, line 59: I think you can remove that and the `spamLinks.txt` file entirely.

src/main/java/net/javadiscord/javabot/systems/moderation/AutoMod.java
```diff
@@ -121,25 +140,33 @@ private void checkContentAutomod(@Nonnull Message message) {
...
-					} catch (Exception e) {
-						e.printStackTrace();
+			while (urlMatcher.find()) {
+				String url = messageRaw.substring(urlMatcher.start(1), urlMatcher.end());
+				try {
+					URI uri = new URI(url);
+					Bot.config.get(message.getGuild()).getModeration().getLogChannel().sendMessage(uri.getHost()).queue();
+					if (spamUrls.contains(uri.getHost())) {
+						if (message.getMember() != null){
+							Bot.config.get(message.getGuild()).getModeration().getLogChannel().sendMessage("Suspicious link by: ".concat("@" + message.getMember().getEffectiveName()).concat(" (" + message.getMember().getId() + ") ").concat(url)).queue();
```
DynxstyGIT, line 155: You could use `Misc.sendToLog(...);` instead

DynxstyGIT (changes requested):

src/main/java/net/javadiscord/javabot/systems/moderation/AutoMod.java
sahilasopa: well, what if due to some exception we were not able to fetch from url, we are still safe to some extend so it might not be a good idea to remove that file

src/main/java/net/javadiscord/javabot/systems/moderation/AutoMod.java
sahilasopa: done

src/main/java/net/javadiscord/javabot/systems/moderation/AutoMod.java
sahilasopa: done

src/main/java/net/javadiscord/javabot/systems/moderation/AutoMod.java
sahilasopa: alright

src/main/java/net/javadiscord/javabot/systems/moderation/AutoMod.java
```diff
@@ -121,25 +140,32 @@ private void checkContentAutomod(@Nonnull Message message) {
...
-										"Automod: Suspicious Link",
-										message.getGuild().getMember(message.getJDA().getSelfUser()),
-										message.getTextChannel(),
-										false
-								);
-						message.delete().queue();
-					} catch (Exception e) {
-						e.printStackTrace();
+			while (urlMatcher.find()) {
+				String url = messageRaw.substring(urlMatcher.start(1), urlMatcher.end());
```
danthe1st, line 149: maybe use `urlMatcher.group(1)` instead of `messageRaw.substring(urlMatcher.start(1), urlMatcher.end())` here.

sahilasopa: won't it only return first url?

danthe1st: https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/regex/Matcher.html#group(int)
> Returns the input subsequence **captured by the given group** during the previous match operation.

PR #161, opened 2022-01-09
Updated the way search results are displayed

MoonTM-GIT:
I have changed the way individual search results are displayed. 

If one exists within the first 320 characters it will now use the last period to end the preview, if there is none it will just do it the way it has been before.

### DIFF
--- a/src/main/java/net/javadiscord/javabot/systems/commands/SearchCommand.java
+++ b/src/main/java/net/javadiscord/javabot/systems/commands/SearchCommand.java
@@ -77,8 +77,13 @@ public class SearchCommand implements SlashCommandHandler {
                 name = object.get("name").getAsString();
                 url = object.get("url").getAsString();
                 snippet = object.get("snippet").getAsString();
-                if (object.get("snippet").getAsString().length() > 260) {
-                    snippet = object.get("snippet").getAsString().substring(0, 260).concat("...");
+                if (object.get("snippet").getAsString().length() > 320) {
+                    snippet = object.get("snippet").getAsString().substring(0, 320);
+                    int snippetLastPeriod = snippet.lastIndexOf('.');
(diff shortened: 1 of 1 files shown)


src/main/java/net/javadiscord/javabot/systems/commands/SearchCommand.java
```diff
@@ -77,8 +77,13 @@ public ReplyAction handle(SlashCommandEvent event) {
                 name = object.get("name").getAsString();
                 url = object.get("url").getAsString();
                 snippet = object.get("snippet").getAsString();
-                if (object.get("snippet").getAsString().length() > 260) {
-                    snippet = object.get("snippet").getAsString().substring(0, 260).concat("...");
+                if (object.get("snippet").getAsString().length() > 320) {
+                    snippet = object.get("snippet").getAsString().substring(0, 320);
```
DynxstyGIT, line 81: Just use `snippet` instead of `object.get("snippet").getAsString()`

src/main/java/net/javadiscord/javabot/systems/commands/SearchCommand.java
```diff
@@ -77,8 +77,13 @@ public ReplyAction handle(SlashCommandEvent event) {
...
+                    int snippetLastPeriod = snippet.lastIndexOf('.');
+                    if (snippetLastPeriod != -1)
+                        snippet = snippet.substring(0, snippetLastPeriod + 1);
+                    else
+                        snippet = snippet.concat("...");
```
DynxstyGIT, line 86: For multi-line if-statements, please add curly brackets. Makes the whole thing more readable

DynxstyGIT (changes requested):

DynxstyGIT (approved):

PR #163, opened 2022-01-11
Added ID to warn embed & Doubled Warn Severities

MoonTM-GIT:
I have noticed that the current embed when a user is warned didn't show the ID which made it necessary to manually look up the warn ID before being able to remove it.

Another thing I think needs to be changed is the "points" the warn severities give, I want this to be discussed - the following is just my opinion;

In my opinion, with warns only counting towards a ban for 30 days being able to get away with 2 *High* severity warns should not be possible. I think doubling the "points" would be a good approach as that allows for two medium and one low warn and, which I think is pretty important, only one high severity warn combined with a low warn before the next warn will result in a ban.


DynxstyGIT:
> I think doubling the "points" would be a good approach
yup, sounds good

DynxstyGIT (approved):

PR #173, opened 2022-01-16
Added /format-code command

MaeveS2:


src/main/java/net/javadiscord/javabot/systems/commands/FormatCodeCommand.java
```diff
@@ -0,0 +1,35 @@
...
+
+import java.util.Objects;
+
+public class FormatCodeCommand implements SlashCommandHandler {
+    @Override
+    public ReplyAction handle(SlashCommandEvent event)  {
+        long id;
+        try {
+            id = Objects.requireNonNull(event.getOption("message-id")).getAsLong();
+        } catch (Exception e) {
```
DynxstyGIT, line 17: I suggest to just check if `event.getOption(...)` is null

src/main/java/net/javadiscord/javabot/systems/commands/FormatCodeCommand.java
```diff
@@ -0,0 +1,35 @@
...
+            if (event.getChannel().hasLatestMessage()) {
+                id = event.getChannel().getLatestMessageIdLong();
+            } else {
+                return Responses.error(event, "Execution failed; no message ID provided and no previous message.");
+            }
+        }
+
+        Message message;
+        try {
+            message = event.getChannel().retrieveMessageById(id).complete();
```
DynxstyGIT, line 27: And don't use `.complete()`

DynxstyGIT (changes requested):

DynxstyGIT (approved):

PR #178, opened 2022-01-22
Starboard Threads

DynxstyGIT:


src/main/java/net/javadiscord/javabot/systems/starboard/StarboardManager.java
```diff
@@ -21,14 +21,14 @@
 	@Override
 	public void onMessageReactionAdd(@NotNull MessageReactionAddEvent event) {
 		if (!validUser(event.getUser())) return;
-		if (event.getChannelType() != ChannelType.TEXT) return;
+		if (event.getChannelType() != ChannelType.TEXT || !event.getChannelType().isThread()) return;
```
danthe1st, line 24: shouldn't that be `&&`?

PR #189, opened 2022-01-29
Fixed bug which caused Starboard listener to repost suspicious Links

### LINKED ISSUE #188: SuggestionListener reposts suspicious links which are deleted by Automod
When a user sends a suspicious link (such as phishing urls) the link gets deleted by automod, but if the same link in sent in #Suggestions SuggestionListener does no find anything wrong and posts it as an embed

To reproduce the same issue simply share a suspicious url in #suggestions and see the magic

Its expected for suggestion listener to delete the message rather then reposting it

sahilasopa:
Closes #188


sahilasopa:
Had to close it, returning values cause bot to crash

Issue #216, opened 2022-02-19
Use modals for #looking-for-programmer 


MoonTM-GIT:
As the quality of posts in the looking-for-programmers channel is very low, we should make creating posts in there only possible using a modal interaction with some basic information.

DynxstyGIT:
I thought we're waiting for the release of forum channels?

MoonTM-GIT:
okaz

PR #219, opened 2022-02-20
Fixed issue with /format-code formatting bot messages

### LINKED ISSUE #215: Make /format-code skip bot messages
Currently, /format-code always formats the LAST message, even if it's sent by the bot itself. This is unnecessary and not what people expect

sahilasopa:
closes #215

### DIFF
--- a/src/main/java/net/javadiscord/javabot/systems/commands/FormatCodeCommand.java
+++ b/src/main/java/net/javadiscord/javabot/systems/commands/FormatCodeCommand.java
@@ -28,6 +28,9 @@ public class FormatCodeCommand implements ISlashCommand, IMessageContextCommand 
 		if (idOption == null) {
 			if (event.getChannel().hasLatestMessage()) {
 				id = event.getChannel().getLatestMessageIdLong();
+				if (event.getChannel().retrieveMessageById(id).complete().getAuthor().isBot()){
+					return Responses.error(event, "Missing required arguments.");
+				}
 			} else {
 				return Responses.error(event, "Missing required arguments.");
 			}
--- a/src/main/java/net/javadiscord/javabot/systems/commands/FormatCodeCommand.java
+++ b/src/main/java/net/javadiscord/javabot/systems/commands/FormatCodeCommand.java
@@ -14,7 +14,6 @@ import net.javadiscord.javabot.command.interfaces.ISlashCommand;
 import net.javadiscord.javabot.data.config.guild.SlashCommandConfig;
 
 import java.time.Instant;
-import java.util.Collections;
 
 /**
  * Command that allows members to format messages.
--- a/src/main/java/net/javadiscord/javabot/systems/commands/FormatCodeCommand.java
+++ b/src/main/java/net/javadiscord/javabot/systems/commands/FormatCodeCommand.java
@@ -29,8 +29,7 @@ public class FormatCodeCommand implements ISlashCommand, IMessageContextCommand 
 		String format = formatOption == null ? "java" : formatOption.getAsString();
 		if (idOption == null) {
 			if (event.getChannel().hasLatestMessage()) {
-				event.getChannel().getHistory().retrievePast(100) // max limit
-						.map(messages -> messages)
+				event.getChannel().getHistory().retrievePast(10)
 						.queue(messages -> {
 							for (Message message : messages) {
 								if (!message.getAuthor().isBot()){
--- a/src/main/java/net/javadiscord/javabot/systems/commands/FormatCodeCommand.java
+++ b/src/main/java/net/javadiscord/javabot/systems/commands/FormatCodeCommand.java
@@ -19,7 +19,7 @@ import java.time.Instant;
  * Command that allows members to format messages.
  */
 public class FormatCodeCommand implements ISlashCommand, IMessageContextCommand {
-	long id;
+	private long id;
 
 	@Override
 	public ReplyCallbackAction handleSlashCommandInteraction(SlashCommandInteractionEvent event) {
--- a/src/main/java/net/javadiscord/javabot/systems/commands/FormatCodeCommand.java
+++ b/src/main/java/net/javadiscord/javabot/systems/commands/FormatCodeCommand.java
@@ -14,33 +14,44 @@ import net.javadiscord.javabot.command.interfaces.ISlashCommand;
 import net.javadiscord.javabot.data.config.guild.SlashCommandConfig;
 
 import java.time.Instant;
+import java.util.Collections;
 
 /**
  * Command that allows members to format messages.
  */
 public class FormatCodeCommand implements ISlashCommand, IMessageContextCommand {
+	long id;
+
 	@Override
 	public ReplyCallbackAction handleSlashCommandInteraction(SlashCommandInteractionEvent event) {
 		var idOption = event.getOption("message-id");
 		var formatOption = event.getOption("format");
+		var slashConfig = Bot.config.get(event.getGuild()).getSlashCommand();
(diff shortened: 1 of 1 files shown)


src/main/java/net/javadiscord/javabot/systems/commands/FormatCodeCommand.java
```diff
@@ -28,6 +28,9 @@ public ReplyCallbackAction handleSlashCommandInteraction(SlashCommandInteraction
 		if (idOption == null) {
 			if (event.getChannel().hasLatestMessage()) {
 				id = event.getChannel().getLatestMessageIdLong();
+				if (event.getChannel().retrieveMessageById(id).complete().getAuthor().isBot()){
+					return Responses.error(event, "Missing required arguments.");
+				}
```
DynxstyGIT, line 33: Don't use `.complete()` calls, as they're blocking the current thread. In addition, I think it would be useful if it takes the last message that isn't sent from a bot instead of throwing an error.

DynxstyGIT (changes requested):

MoonTM-GIT (changes requested):
Yeah, this does not add the functionality that was described in #215.

src/main/java/net/javadiscord/javabot/systems/commands/FormatCodeCommand.java
sahilasopa: fetch last 5 messages and format latest non bot message?

MoonTM-GIT: for example, yes

MoonTM-GIT (approved):

src/main/java/net/javadiscord/javabot/systems/commands/FormatCodeCommand.java
```diff
@@ -19,25 +19,38 @@
...
 	public ReplyCallbackAction handleSlashCommandInteraction(SlashCommandInteractionEvent event) {
 		var idOption = event.getOption("message-id");
 		var formatOption = event.getOption("format");
+		var slashConfig = Bot.config.get(event.getGuild()).getSlashCommand();
 		String format = formatOption == null ? "java" : formatOption.getAsString();
-		long id;
 		if (idOption == null) {
 			if (event.getChannel().hasLatestMessage()) {
-				id = event.getChannel().getLatestMessageIdLong();
+				event.getChannel().getHistory().retrievePast(100) // max limit
```
danthe1st, line 32: I think fetching 100 messages is a little bit of an overkill.

src/main/java/net/javadiscord/javabot/systems/commands/FormatCodeCommand.java
```diff
@@ -19,25 +19,38 @@
...
+						.map(messages -> messages)
+						.queue(messages -> {
+							for (Message message : messages) {
+								if (!message.getAuthor().isBot()){
+									id = message.getIdLong();
+									break;
+								}
+							}
+							event.getTextChannel().retrieveMessageById(id).queue(
```
danthe1st, line 41: `id` may be `null` if all the messages were sent by the bot (could happen in channels like the suggestion channel or looking-for-programmer if #216 is implemented and deployed)

src/main/java/net/javadiscord/javabot/systems/commands/FormatCodeCommand.java
```diff
@@ -19,25 +19,38 @@
...
 		var idOption = event.getOption("message-id");
 		var formatOption = event.getOption("format");
+		var slashConfig = Bot.config.get(event.getGuild()).getSlashCommand();
 		String format = formatOption == null ? "java" : formatOption.getAsString();
-		long id;
 		if (idOption == null) {
 			if (event.getChannel().hasLatestMessage()) {
-				id = event.getChannel().getLatestMessageIdLong();
+				event.getChannel().getHistory().retrievePast(100) // max limit
+						.map(messages -> messages)
```
danthe1st, line 33: Does this have any purpose?

src/main/java/net/javadiscord/javabot/systems/commands/FormatCodeCommand.java
sahilasopa: how many do you think we should fetch? 10?

sahilasopa: how many do you think we should fetch? 10?

src/main/java/net/javadiscord/javabot/systems/commands/FormatCodeCommand.java
sahilasopa: alright imma add a null check

MoonTM-GIT:
You didn't add that null-check

MoonTM-GIT (approved):

src/main/java/net/javadiscord/javabot/systems/commands/FormatCodeCommand.java
MoonTM-GIT: Yeah, you're right

danthe1st:
> You didn't add that null-check

They have an action in case the request fails but I think that invoking `retrieveMessageById` with a `0` id throws an `IllegalArgumentException` when synchronously.

sahilasopa:
> > You didn't add that null-check
> 
> They have an action in case the request fails but I think that invoking `retrieveMessageById` with a `0` id throws an `IllegalArgumentException` when synchronously.

@danthe1st I tried multiple things to test this, including using this command in empty channels I was not able to reproduce the same errer, can you please tell me when exactly is it thrown?

src/main/java/net/javadiscord/javabot/systems/commands/FormatCodeCommand.java
sahilasopa: it is handled actually
`e -> Responses.error(event.getHook(), "Could not retrieve message.").queue())`

MoonTM-GIT:
> > You didn't add that null-check
> 
> They have an action in case the request fails but I think that invoking `retrieveMessageById` with a `0` id throws an `IllegalArgumentException` when synchronously.

Yup, I just looked at the changes of that one commit

src/main/java/net/javadiscord/javabot/systems/commands/FormatCodeCommand.java
```diff
@@ -19,25 +19,37 @@
  * Command that allows members to format messages.
  */
 public class FormatCodeCommand implements ISlashCommand, IMessageContextCommand {
+	long id;
```
DynxstyGIT, line 22: Why is `id` a class variable

DynxstyGIT (changes requested):

src/main/java/net/javadiscord/javabot/systems/commands/FormatCodeCommand.java
sahilasopa: i will make this private, it is a class variable because its being used in lambda expressions and the require variables to be final or atomic, to keep things simple it was better to make it a class variable

danthe1st: Just create the `id` variable only in the blpcks you need it (`queue` and `else`)?

sahilasopa: this wont work as variables are used both in and outside lambda functions (As Far As I Know)

DynxstyGIT (approved):

PR #220, opened 2022-02-20
fix data corruption caused by duplicate commands

danthe1st:
# Bug fixed
When creating a slash command with the same name as an existing command, it causes a data corruption.
This results in an exception both at that time and on startup.
Other custom commands may not be registered because of this.

image

# Fix
This bug is fixed by checking whether a command with the same name already exists and responds with an error if so.
In order to achieve this, a new public method `doesSlashCommandExist(String,Guild)` (and a new private method `getCustomCommand`) are created in `InteractionHandler`.

# Alternatives considered
### Use of a transaction
If a transaction was used for creating the custom command (auto-commit disabled), an exception (in `InteractionHandler#registerCommands`) caused by the duplicate command would cause a rollback.

However, the exception would not be thrown as `InteractionHandler#registerCommands` uses a different `Connection` object so it the changes to the custom commands would not be applied resulting in no exception and no update to the custom command when there are no conflicts.

### `InteractionHandler#doesGlobalSlashCommandExist(String)`
Instead of creating a method `doesSlashCommandExist` that checks if a slash command with a given name exists in a guild, a method checking if a global, non-custom slash command existed or even return that command.

However, this would expose the existence handling of custom commands in `InteractionHandler` to other classes.

### DIFF
--- a/src/main/java/net/javadiscord/javabot/command/InteractionHandler.java
+++ b/src/main/java/net/javadiscord/javabot/command/InteractionHandler.java
@@ -79,6 +79,12 @@ public class InteractionHandler extends ListenerAdapter {
 		}
 	}
 
+	/**
+	 * Checks whether a slash command exists
+	 * @param name the name of the command
+	 * @param guild the {@link Guild} the command may exist in
+	 * @return <code>true</code> if the command exists, else <code>false</code>
+	 */
 	public boolean doesSlashCommandExist(String name, Guild guild){
 		try {
 			return this.slashCommandIndex.containsKey(name) || getCustomCommand(name, guild).isPresent();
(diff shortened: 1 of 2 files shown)


MoonTM-GIT (approved):

DynxstyGIT (approved):

MoonTM-GIT (changes requested):
Could you please fix the checkstyle problems?

danthe1st:
I already fixed them in c6d515160f1c22be39e1ed05903013e64c4484ec

MoonTM-GIT (approved):

danthe1st:
The checkstyle problems are actually fixed in e1852ec2f6e5c3870da5ec1efd830c367e35acfd (see the CI build for details).

MoonTM-GIT:
Yup, looks good. Thank you

PR #233, opened 2022-03-06
Inverted Condition

DeveloperCallum:
Fix the inverted Condition was `"CommandUser role bigger than target role"`, and is now `"CommandUser role less than or equal to target role"`

### DIFF
--- a/src/main/java/net/javadiscord/javabot/command/moderation/ModerateUserCommand.java
+++ b/src/main/java/net/javadiscord/javabot/command/moderation/ModerateUserCommand.java
@@ -44,14 +44,16 @@ public abstract class ModerateUserCommand extends ModerateCommand {
 
 		if (target.isOwner()) {
 			return Responses.error(event, "You cannot preform actions on a higher member staff member.");
-		}
-
-		//If both users have at least one role.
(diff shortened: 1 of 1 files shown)


src/main/java/net/javadiscord/javabot/command/moderation/ModerateUserCommand.java
```diff
@@ -44,15 +44,17 @@ protected final ReplyCallbackAction handleModerationCommand(SlashCommandInteract
 
 		if (target.isOwner()) {
 			return Responses.error(event, "You cannot preform actions on a higher member staff member.");
-		}
-
-		//If both users have at least one role.
-		if (target.getRoles().size() > 0 && commandUser.getRoles().size() > 0) {
-			if (commandUser.getRoles().get(0).getPosition() > target.getRoles().get(0).getPosition()) {
-				return Responses.error(event, "You cannot preform actions on a higher member staff member.");
+		}else{
```
DynxstyGIT, line 47:
```suggestion
		} else {
```

DynxstyGIT (changes requested):

DynxstyGIT (approved):

PR #234, opened 2022-03-07
Upgraded to JDA v5.0.0-alpha.9 (Xirado's Fork for Modal Support)

DynxstyGIT:


src/main/java/net/javadiscord/javabot/systems/moderation/ReportCommand.java
```diff
@@ -119,7 +119,7 @@ private void handleMessageReport(ModalInteractionEvent event, String messageId)
 
 	private void handleUserReport(ModalInteractionEvent event, String userId) {
 		String reason = event.getValue(REASON_OPTION_NAME).getAsString();
-		if(reason.isBlank()) {
+		if (reason.isBlank()) {
```
MoonTM-GIT, line 122: not a fan of this but whatever

MoonTM-GIT (approved):

PR #235, opened 2022-03-10
Refactored Report Messages to use Threads

### LINKED ISSUE #230: Set reports as handled
**Is your feature request related to a problem? Please describe.**
It currently is impossable to tell if a report  has been handled by another user. this could cause multiple staff to handle the same request.

**Describe the solution you'd like**
if someone handles a report the embed is updated to say REPORT HANDLED BY X STAFF.

**Describe alternatives you've considered**
None.

DynxstyGIT:
This PR adds a thread to each report message which can be used to discuss further actions.
This also introduces a "Mark as resolved" button (Closes #230), which simply archives the thread.

image

image

image


MoonTM-GIT:
Yeah, I think this is better than the current system.

PR #238, opened 2022-03-12
Improved Suggestions

DynxstyGIT:
This PR improves suggestions quite a bit by automatically creating thread channels for new suggestions and adding a new "On Hold" state for suggestions.

image


MoonTM-GIT (approved):

MoonTM-GIT:
I think another state for suggestions that are already implemented would be really helpful

DynxstyGIT:
Or we'll add a `reason` option to the decline subcommand with some predefined options/choiches

MoonTM-GIT:
As long as it's optional I'm fine with it

PR #239, opened 2022-03-12
Ban Command Upgrades

DeveloperCallum:


src/main/java/net/javadiscord/javabot/systems/moderation/ModerationService.java
```diff
@@ -198,21 +201,43 @@ public boolean removeTimeout(Member member, String reason, Member removedBy, Mes
...
+	/**
+	 * Bans a member.
+	 *
+	 * @param snowFlakeID The member to ban.
+	 * @param reason      The reason for banning the member.
+	 * @param bannedBy    The member who is responsible for banning this member.
+	 * @param channel     The channel in which the ban was issued.
+	 * @param quiet       If true, don't send a message in the channel.
+	 */
+	public void ban(String snowFlakeID, String reason, Member bannedBy, MessageChannel channel, boolean quiet) throws IllegalArgumentException, InsufficientPermissionException, HierarchyException {
```
DynxstyGIT, line 235: maybe rename `snowFlakeID` to `targetId` and make it a long instead of a String

src/main/java/net/javadiscord/javabot/systems/moderation/ModerationService.java
```diff
@@ -380,6 +406,18 @@ private MessageEmbed buildBanEmbed(Member member, String reason, Member bannedBy
...
+		return new EmbedBuilder()
+				.setColor(Color.RED)
+				.setTitle(String.format("%s | Ban", snowflakeID))
+				.addField("User", snowflakeID, true)
+				.addField("Banned by", bannedBy.getAsMention(), true)
+				.addField("Ban Reason", String.format(reasonFormat, reason), false)
+				.setTimestamp(Instant.now())
+				.setFooter(bannedBy.getUser().getAsTag(), bannedBy.getEffectiveAvatarUrl())
+				.build();
+	}
```
DynxstyGIT, line 419: Same as above. Rename `snowFlakeID` and make it a long instead. 
+ Instead of `Color.RED` use the ErrorColor from the SlashCommandConfig (GuildConfig#getSlashCommand)

src/main/java/net/javadiscord/javabot/systems/moderation/ModerationService.java
```diff
@@ -198,21 +201,43 @@ public boolean removeTimeout(Member member, String reason, Member removedBy, Mes
...
+			} catch (InsufficientPermissionException | HierarchyException e) {
+				throw new UserPermissionException(channel.getJDA().getSelfUser(), e);
+			}
+
 			config.getLogChannel().sendMessageEmbeds(banEmbed).queue();
 			if (!quiet) channel.sendMessageEmbeds(banEmbed).queue();
-			return true;
+		} else {
+			throw new UserPermissionException(bannedBy.getUser(), Permission.BAN_MEMBERS, "User does not have permissions to ban the targeted member.");
 		}
```
DynxstyGIT, line 223: Just invert the if statement and put the Exception in there

src/main/java/net/javadiscord/javabot/systems/moderation/ModerationService.java
```diff
@@ -198,21 +201,43 @@ public boolean removeTimeout(Member member, String reason, Member removedBy, Mes
...
 					c -> c.sendMessage(config.getBanMessageText()).setEmbeds(banEmbed).queue(),
 					e -> log.info("Could not send Ban Direct Message to User {}", member.getUser().getAsTag())
 			);
-			bannedBy.getGuild().ban(member, BAN_DELETE_DAYS, reason).queue();
+
+			try {
+				bannedBy.getGuild().ban(member, BAN_DELETE_DAYS, reason).queue();
+			} catch (InsufficientPermissionException | HierarchyException e) {
+				throw new UserPermissionException(channel.getJDA().getSelfUser(), e);
+			}
```
DynxstyGIT, line 217: This does not throw any exceptions. You would need to catch the exception in the queue's lambda

src/main/java/net/javadiscord/javabot/systems/moderation/ModerationService.java
```diff
@@ -198,21 +201,43 @@ public boolean removeTimeout(Member member, String reason, Member removedBy, Mes
 	 * @param bannedBy The member who is responsible for banning this member.
 	 * @param channel  The channel in which the ban was issued.
 	 * @param quiet    If true, don't send a message in the channel.
-	 * @return Whether the moderator has the permission to ban this member or not.
 	 */
-	public boolean ban(Member member, String reason, Member bannedBy, MessageChannel channel, boolean quiet) {
+	public void ban(Member member, String reason, Member bannedBy, MessageChannel channel, boolean quiet) throws IllegalArgumentException, UserPermissionException {
 		var banEmbed = buildBanEmbed(member, reason, bannedBy);
 		if (canBanUser(member, bannedBy)) {
```
DynxstyGIT, line 207: Just do `if(!canBanUser(...)) { throw ... }`

src/main/java/net/javadiscord/javabot/systems/moderation/BanCommand.java
```diff
@@ -6,27 +6,68 @@
...
+		try {
+			moderationService.ban(snowflakeID, reason, commandUser, event.getTextChannel(), quiet);
+		} catch (IllegalArgumentException e) {
+			return Responses.error(event, "Incorrect Argument");
+		} catch (UserPermissionException e) {
+			return Responses.error(event, e.toString());
+		}
+
+		return Responses.success(event, "User Banned", String.format("%s has been banned.", snowflakeID));
+	}
```
DynxstyGIT, line 58: Just keep this method (or simply merge it with the handleModerationCommand method above) and remove the other one. There's no need to have an extra method for a member object if we can just use `Member#getIdLong`
However, rename the `snowflakeID` to `targetId` and make it a long instead of a String.

src/main/java/net/javadiscord/javabot/command/moderation/UserPermissionException.java
```diff
@@ -0,0 +1,47 @@
...
+	@Override
+	public String getMessage() {
+		return super.getMessage();
+	}
+
+	@Override
+	public String toString() {
+		return getUser().getName() + " " + super.getMessage();
+	}
+}
```
DynxstyGIT, line 47: Do we need this in general? If there needs an exception to be thrown, we could simply use the ResponseException class.
And as described in the other comments down below, you cannot catch a `HierachyException` or `InsufficentPermissionException` from a queue call

DynxstyGIT (changes requested):

src/main/java/net/javadiscord/javabot/systems/moderation/ModerationService.java
DeveloperCallum: That is how the oringinal command was and its not for an error so why use the error colour?

DeveloperCallum: The reason it is a String is beacuse A:
- Discord IDs do not specify that they are long numbers.
- Discord will accept any "number" in the user portion `i.e 0000034243234`
- Ban does not have an option for long, it does for string. 
image

src/main/java/net/javadiscord/javabot/systems/moderation/ModerationService.java
DeveloperCallum: Yes, i could have used a gaurd statement.

src/main/java/net/javadiscord/javabot/systems/moderation/ModerationService.java
DynxstyGIT: Discord IDs are by default always long numbers. 
JDA kinda defaults it to Strings tho, as `getId` returns a String and `getIdLong` a Long.
Other libraries like javacord do it differently and use longs as the default id, idk.
However, we're using long numbers as ids everywhere so why break the code style and use string instead if it is not necessary.  And I'm pretty sure you can use longs in the ban method. And if not this may be a worthy PR for JDA, cuz they allow basically everywhere to use either strings or longs as user ids.

DynxstyGIT: And the Error Color is used in many different places. It's just a red-ish color. It's not exclusivly meant for errors

src/main/java/net/javadiscord/javabot/systems/moderation/BanCommand.java
DeveloperCallum: This method does not check for much since it's using the ID and has no garantee that the user exists within the server whereas, member does. the user option makes sure that the user is in the server (if it returns a member)

MoonTM-GIT: The entire point of this PR is to get rid of that check, no?

src/main/java/net/javadiscord/javabot/systems/moderation/ModerationService.java
DeveloperCallum: as stated in an earlier comment, "Ban does not have an option for long, it does for string." Cconverting that to whatever, so it can be used as long in other calls just to turn it back would be meaningless.

src/main/java/net/javadiscord/javabot/systems/moderation/ModerationService.java
DeveloperCallum: No, as explained in the comment below. 

https://github.com/Java-Discord/JavaBot/pull/239#discussion_r825337024

DynxstyGIT:
I'll close this PR for now, as it's kinda outdated. I'll rework the ban system myself in the near future

PR #243, opened 2022-03-23
add user when creating suggestion thread

danthe1st:
This Pull Request adds a the user to the corresponding thread when creating a submission.
Furthermore, it refactors the addition of attachments to suggestion so it doesn't block the JDA listener thread.

### DIFF
--- a/src/main/java/net/javadiscord/javabot/events/SuggestionListener.java
+++ b/src/main/java/net/javadiscord/javabot/events/SuggestionListener.java
@@ -15,7 +15,9 @@ import net.javadiscord.javabot.data.config.guild.SlashCommandConfig;
 import org.jetbrains.annotations.NotNull;
 
 import java.time.Instant;
-import java.util.concurrent.ExecutionException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
 
 /**
  * Listens for {@link MessageReceivedEvent}s in the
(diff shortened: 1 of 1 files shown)


src/main/java/net/javadiscord/javabot/events/SuggestionListener.java
```diff
@@ -32,14 +34,17 @@ public void onMessageReceived(@NotNull MessageReceivedEvent event) {
 		}
 		var config = Bot.config.get(event.getGuild());
 		MessageEmbed embed = this.buildSuggestionEmbed(event.getMessage(), config.getSlashCommand());
-		this.addAttachments(event.getMessage(), event.getChannel().sendMessageEmbeds(embed)).queue(message -> {
+		this.addAttachmentsAndSend(event.getMessage(), event.getChannel().sendMessageEmbeds(embed)).thenAccept(message -> {
 					this.addReactions(message).queue();
```
Kemikals, line 38: should probably put this block into a method to properly describe what's happening, also odd use of this everywhere

danthe1st: What do you mean with `odd use of this`? Aside from that, what block are you talking about? The lambda body?

src/main/java/net/javadiscord/javabot/events/SuggestionListener.java
```diff
@@ -68,21 +75,23 @@ private boolean canCreateSuggestion(MessageReceivedEvent event) {
...
-	 * Adds all Attachments from the initial message to the new message action.
+	 * Adds all Attachments from the initial message to the new message action and sends the message.
 	 *
 	 * @param message The initial {@link Message} object.
 	 * @param action The new {@link MessageAction}.
-	 * @return The complete {@link MessageAction} with all attachments.
+	 * @return A {@link CompletableFuture} with the message that is being sent.
 	 */
-	private MessageAction addAttachments(Message message, MessageAction action) {
+	private CompletableFuture<Message> addAttachmentsAndSend(Message message, MessageAction action) {
```
Kemikals, line 84: Moving the loop logic into a method, and using a stream to collect cleans this up a lot. 

`List<CompletableFuture<?>> attachmentFutures = message.getAttachments().stream()
		.map(attachment -> handleAttachment(action, attachment))
		.collect(Collectors.toList());`

src/main/java/net/javadiscord/javabot/events/SuggestionListener.java
Kemikals: yes, use methods to encapsulate that nested logic. For the this there's methods calling it prefixed with this some places, other places not. It's not a necessary call but if were going to use it that way should be standardized

src/main/java/net/javadiscord/javabot/events/SuggestionListener.java
danthe1st: To be honest, moving this code to a method isn't worth it in my opinion since the code is pretty obvious.

src/main/java/net/javadiscord/javabot/events/SuggestionListener.java
danthe1st: The `this` here is not context of this PR as I just left it as-is.

MoonTM-GIT (approved):

PR #255, opened 2022-04-13
Help Channel Experience

DynxstyGIT:


src/main/java/net/javadiscord/javabot/systems/help/HelpChannelManager.java
```diff
@@ -319,6 +319,15 @@ private void unreserveChannelByOtherUser(TextChannel channel, User owner, @Nulla
 	public RestAction<?> unreserveChannel(TextChannel channel) {
 		if (this.config.isRecycleChannels()) {
 			try (var con = Bot.dataSource.getConnection()) {
+				HelpExperienceService service = new HelpExperienceService(Bot.dataSource);
+				Optional<ChannelReservation> reservationOptional = this.getReservationForChannel(channel.getIdLong());
+				if (reservationOptional.isPresent()) {
+					ChannelReservation reservation = reservationOptional.get();
+					Map<Long, Double> experience = this.calculateExperience(HelpChannelListener.reservationMessages.get(reservation), reservation.getUserId());
+					for (Long recipient : experience.keySet()) {
+						service.performTransaction(recipient, experience.get(recipient), String.format("Helped <@%s> in <@%s>", reservation.getUserId(), channel.getIdLong()));
```
danthe1st, line 328: probably use a enums and save user ids instead of using `String`s

src/main/java/net/javadiscord/javabot/systems/help/HelpExperienceJob.java
```diff
@@ -0,0 +1,18 @@
...
+import org.quartz.JobExecutionContext;
+import org.quartz.JobExecutionException;
+
+/**
+ * Removes a specified amount of experience from everyone's help account.
+ */
+public class HelpExperienceJob extends DiscordApiJob {
+	@Override
+	protected void execute(JobExecutionContext context, JDA jda) throws JobExecutionException {
+		DbHelper.doDaoAction(HelpAccountRepository::new, dao -> dao.removeExperienceFromAllAccounts(-1));
```
danthe1st, line 16: I think that should be made configurable as well.

src/main/java/net/javadiscord/javabot/systems/help/dao/HelpAccountRepository.java
```diff
@@ -0,0 +1,90 @@
...
+	private final Connection con;
+
+	/**
+	 * Inserts a new {@link HelpAccount}.
+	 *
+	 * @param account The account that should be inserted.
+	 * @throws SQLException If an error occurs.
+	 */
+	public void insert(HelpAccount account) throws SQLException {
+		try (PreparedStatement s = con.prepareStatement("INSERT INTO help_account (user_id, experience) VALUES ( ?, ? )")) {
```
danthe1st, line 28: Do we really need that table? Couldn't we just use the sum of all experiences of that user in the transaction table?

src/main/java/net/javadiscord/javabot/systems/help/model/HelpTransaction.java
```diff
@@ -0,0 +1,19 @@
...
+ * Data class that represents a single Help Transaction.
+ */
+@Data
+public class HelpTransaction {
+	private long id;
+	private long recipient;
+	private LocalDateTime createdAt;
+	private double value;
+	@Nullable
+	private String message;
```
danthe1st, line 18: As stated previously, I don't think that message is necessary. Instead we could use an enum for the type of transaction and a user id of the user causing the transaction.The message could then be derived when necessary (from the transaction type, )

src/main/resources/migrations/04-13-2022_add_help_xp_tables.sql
```diff
@@ -0,0 +1,12 @@
...
+  user_id BIGINT PRIMARY KEY,
+  experience DOUBLE NOT NULL,
+);
+
+CREATE TABLE help_transaction (
+  id BIGINT PRIMARY KEY AUTO_INCREMENT,
+  recipient BIGINT NOT NULL,
+  created_at TIMESTAMP(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6),
+  value DOUBLE NOT NULL,
+  message VARCHAR(255)
```
danthe1st, line 11: I am quite unsure if we really want to store up to 255 characters per transaction as there would be a pretty high amount of those.

src/main/resources/schema.sql
```diff
@@ -155,3 +155,16 @@ CREATE TABLE qotw_submissions (
...
+	  user_id BIGINT PRIMARY KEY,
+	  experience DOUBLE NOT NULL,
+);
+
+CREATE TABLE help_transaction (
+	  id BIGINT PRIMARY KEY AUTO_INCREMENT,
+	  recipient BIGINT NOT NULL,
+	  created_at TIMESTAMP(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6),
+	  value DOUBLE NOT NULL,
+	  message VARCHAR(255)
```
danthe1st, line 169: I am quite unsure if we really want to store up to 255 characters per transaction as there would be a pretty high amount of those.

PR #275, opened 2022-05-02
Moved leaderboards to single command

MoonTM-GIT:
This PR moves all leaderboards to a single command to make it a bit easier to find them. I have not made any changes to the commands, I just moved them


src/main/resources/commands/slash/user.yaml
```diff
@@ -185,4 +180,23 @@
...
+    - description: Question-Of-The-Week Leaderboard.
+      name: qotw
+
+    # /leaderboard thanks
+    - description: Thanks leaderboard.
+      name: thanks
+
+    # /leaderboard help-xp
+    - description: Help Experience Leaderboard.
+      name: help-xp
```
danthe1st, line 200:
```suggestion
      name: help-xp
      options:
        - name: page
          description: The page of results to show. By default it starts at 1.
          type: INTEGER
          required: false
```

danthe1st (changes requested):
This PR wrongly adds a change in functionality.
The `page` argument from the `help-xp` subcommand is removed but still used in the subcommand handler.
The `page` option should be kept in the leaderboard command configuration file.

danthe1st (approved):

DynxstyGIT (approved):

PR #280, opened 2022-05-07
Added name check and change system

### LINKED ISSUE #279: Generate random names for people with non-ascii names
We should generate names for people that do not have an easily pingable nickname. This should only be done if neither their nick- nor their username is pingable.

MoonTM-GIT:
I have added a name check and change system as described in #279. All names that contain an ASCII character not between 32 and 127 (excluding) in the first three characters of their name get a random name.

Names are generated from a list of 1525 English nouns, 1604 Adjectives, and a number from 1 to 9999. If I'm not mistaken (it's been a while since I needed this the last time) this means there are 24,458,553,900 Combinations and a chance of ~0.0000000041% to get a duplicate at the beginning, with the probability obviously going up as names are generated.

Closes #279

### DIFF
--- a/src/main/java/net/javadiscord/javabot/data/h2db/message_cache/MessageCacheListener.java
+++ b/src/main/java/net/javadiscord/javabot/data/h2db/message_cache/MessageCacheListener.java
@@ -67,8 +67,9 @@ public class MessageCacheListener extends ListenerAdapter {
 	 * @return true if any of the criteria above apply
 	 */
 	private boolean ignoreMessageCache(Message message) {
+		if (!message.isFromGuild()) return true;
 		MessageCacheConfig config = Bot.config.get(message.getGuild()).getMessageCache();
-		return !message.isFromGuild() || message.getAuthor().isBot() || message.getAuthor().isSystem() ||
+		return message.getAuthor().isBot() || message.getAuthor().isSystem() ||
 				config.getExcludedUsers().contains(message.getAuthor().getIdLong()) ||
 				config.getExcludedChannels().contains(message.getChannel().getIdLong());
 	}
--- a/src/main/java/net/javadiscord/javabot/listener/PingableNameListener.java
+++ b/src/main/java/net/javadiscord/javabot/listener/PingableNameListener.java
@@ -106,8 +106,8 @@ public class PingableNameListener extends ListenerAdapter {
 	 */
 	private static List<String> readStrings(String url) {
 		List<String> list;
-		try (Scanner scan = new Scanner(new URL(url).openStream()).useDelimiter("\\A")) {
-			list = Arrays.stream(scan.next().split("\n")).collect(Collectors.toList());
+		try (Scanner scan = new Scanner(new URL(url).openStream()).useDelimiter("\\n")) {
+			list = scan.tokens().collect(Collectors.toList());
 		} catch (IOException e) {
 			log.error("Error during retrieval of words.");
 			list = new ArrayList<>();
--- a/src/main/java/net/javadiscord/javabot/listener/PingableNameListener.java
+++ b/src/main/java/net/javadiscord/javabot/listener/PingableNameListener.java
@@ -1,15 +1,11 @@
 package net.javadiscord.javabot.listener;
 
 import lombok.extern.slf4j.Slf4j;
-import net.dv8tion.jda.api.Permission;
 import net.dv8tion.jda.api.entities.Member;
-import net.dv8tion.jda.api.entities.TextChannel;
 import net.dv8tion.jda.api.events.guild.member.GuildMemberJoinEvent;
 import net.dv8tion.jda.api.events.guild.member.update.GuildMemberUpdateNicknameEvent;
 import net.dv8tion.jda.api.hooks.ListenerAdapter;
-import net.javadiscord.javabot.Bot;
 import net.javadiscord.javabot.systems.notification.GuildNotificationService;
-import net.javadiscord.javabot.util.GuildUtils;
 import net.javadiscord.javabot.util.StringUtils;
 
 import java.io.IOException;
@@ -41,15 +37,22 @@ public class PingableNameListener extends ListenerAdapter {
 
 	@Override
 	public void onGuildMemberJoin(GuildMemberJoinEvent event) {
-		if (!isPingable(event.getUser().getName())) {
-			changeName(event.getMember());
-		}
+		checkNickname(event.getMember(), null);
 	}
 
 	@Override
 	public void onGuildMemberUpdateNickname(GuildMemberUpdateNicknameEvent event) {
-		if (!isPingable(event.getNewNickname()) && !isPingable(event.getUser().getName())) {
-			changeName(event.getMember());
+		checkNickname(event.getMember(), event.getNewNickname());
+	}
+
+	/**
+	 * Checks whether the given {@link Member}'s nickname should be changed.
+	 * @param member The {@link Member} to check.
+	 * @param nickname The {@link Member}'s new Nickname, null if that does not exist.
+	 */
+	private void checkNickname(Member member, String nickname) {
+		if (!(nickname==null||isPingable(nickname)) && !isPingable(member.getUser().getName())) {
+			changeName(member);
 		}
 	}
 
--- a/src/main/java/net/javadiscord/javabot/listener/PingableNameListener.java
+++ b/src/main/java/net/javadiscord/javabot/listener/PingableNameListener.java
@@ -1,7 +1,6 @@
 package net.javadiscord.javabot.listener;
 
 import lombok.extern.slf4j.Slf4j;
-import net.dv8tion.jda.api.Permission;
 import net.dv8tion.jda.api.entities.Member;
 import net.dv8tion.jda.api.events.guild.member.GuildMemberJoinEvent;
 import net.dv8tion.jda.api.events.guild.member.update.GuildMemberUpdateNicknameEvent;
--- a/src/main/java/net/javadiscord/javabot/listener/PingableNameListener.java
+++ b/src/main/java/net/javadiscord/javabot/listener/PingableNameListener.java
@@ -8,6 +8,7 @@ import net.dv8tion.jda.api.events.guild.member.GuildMemberJoinEvent;
 import net.dv8tion.jda.api.events.guild.member.update.GuildMemberUpdateNicknameEvent;
 import net.dv8tion.jda.api.hooks.ListenerAdapter;
 import net.javadiscord.javabot.Bot;
+import net.javadiscord.javabot.systems.notification.GuildNotificationService;
 import net.javadiscord.javabot.util.GuildUtils;
 import net.javadiscord.javabot.util.StringUtils;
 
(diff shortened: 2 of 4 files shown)


src/main/java/net/javadiscord/javabot/listener/PingableNameListener.java
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	@Override
+	public void onGuildMemberUpdateNickname(GuildMemberUpdateNicknameEvent event) {
+		if (!isPingable(event.getNewNickname()) && !isPingable(event.getUser().getName()) && !canBypassCheck(event.getMember())) {
+			String newName = generateRandomName();
+			event.getMember().modifyNickname(newName).queue();
+			event.getUser().openPrivateChannel()
+					.flatMap(channel -> channel.sendMessageFormat("Your nickname has been set to `%s` since both your user- and nickname's first three characters were deemed as not-pingable.", newName))
+					.queue();
```
DynxstyGIT, line 56: Move this into a separate method

DynxstyGIT: And maybe send a message to the Log Channel

src/main/java/net/javadiscord/javabot/listener/PingableNameListener.java
```diff
@@ -0,0 +1,115 @@
...
+		return true;
+	}
+
+	/**
+	 * Generates a random name consisting of a noun, an adjective and a number from 1 to 9999.
+	 * @return The generated name.
+	 */
+	private String generateRandomName() {
+		String noun = this.nouns.get(random.nextInt(nouns.size()));
+		String adjective = this.adjectives.get(random.nextInt(adjectives.size()));
```
DynxstyGIT, line 83:
```suggestion
		String noun = nouns.get(random.nextInt(nouns.size()));
		String adjective = adjectives.get(random.nextInt(adjectives.size()));
```

src/main/java/net/javadiscord/javabot/listener/PingableNameListener.java
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	/**
+	 * Reads strings from the given URL to a {@link List}, removes words with a dash.
+	 * @param url The URL to read.
+	 * @return A {@link List} of Strings.
+	 */
+	private static List<String> readStrings(String url) {
+		List<String> list;
+		try(Scanner scan = new Scanner(new URL(url).openStream()).useDelimiter("\\A")) {
```
DynxstyGIT, line 95:
```suggestion
		try (Scanner scan = new Scanner(new URL(url).openStream()).useDelimiter("\\A")) {
```

src/main/java/net/javadiscord/javabot/listener/PingableNameListener.java
```diff
@@ -0,0 +1,115 @@
...
+			String response = scan.next();
+			list = Arrays.stream(response.split("\n")).collect(Collectors.toList());
+		} catch (IOException e) {
+			e.printStackTrace();
```
DynxstyGIT, line 99: Use a proper logger

src/main/java/net/javadiscord/javabot/data/h2db/message_cache/MessageCacheListener.java
```diff
@@ -67,8 +67,9 @@ public void onMessageDelete(@NotNull MessageDeleteEvent event) {
 	 * @return true if any of the criteria above apply
 	 */
 	private boolean ignoreMessageCache(Message message) {
+		if (!message.isFromGuild()) return true;
 		MessageCacheConfig config = Bot.config.get(message.getGuild()).getMessageCache();
-		return !message.isFromGuild() || message.getAuthor().isBot() || message.getAuthor().isSystem() ||
+		return message.getAuthor().isBot() || message.getAuthor().isSystem() ||
```
MoonTM-GIT, line 72: This was done because `message#getGuild` throws an exception if the message was not sent in a guild. This would happen when direct messages are sent/received

src/main/java/net/javadiscord/javabot/listener/PingableNameListener.java
```diff
@@ -0,0 +1,115 @@
...
+		list.removeIf(word -> word.contains("-"));
+		return list;
+	}
+
+	/**
+	 * Checks if the given {@link Member} can bypass the name-check.
+	 * @param member The {@link Member} to check.
+	 * @return Whether the Member can bypass name-checks or not.
+	 */
+	private static boolean canBypassCheck(Member member) {
```
danthe1st, line 112: Do we need that? To be honest, I think this should apply for everyone.

DynxstyGIT (changes requested):

src/main/java/net/javadiscord/javabot/listener/PingableNameListener.java
MoonTM-GIT: This is how it's done almost everywhere where a user does not need to be notified. 
image

DynxstyGIT: I know, just use Slf4j#log#error instead tho

src/main/java/net/javadiscord/javabot/listener/PingableNameListener.java
```diff
@@ -0,0 +1,115 @@
...
+
+	/**
+	 * Checks if the given name's first three characters contain ASCII characters not between 32 and 126.
+	 * @param name The name to check.
+	 * @return True if first three characters contain invalid characters, False if not.
+	 */
+	private boolean isPingable(String name) {
+		if (name == null) return true;
+		char[] nameChars = name.toCharArray();
+		for (int i = 0; i < 2; i++) {
```
danthe1st, line 68: This throws a NPE if the name has less than 3 characters.
```suggestion
		for (int i = 0; i < Math.min(2,name.length()); i++) {
```

src/main/java/net/javadiscord/javabot/listener/PingableNameListener.java
MoonTM-GIT: Can't you just add this to checkstyle...

src/main/java/net/javadiscord/javabot/listener/PingableNameListener.java
```diff
@@ -0,0 +1,115 @@
...
+			}
+		}
+		return true;
+	}
+
+	/**
+	 * Generates a random name consisting of a noun, an adjective and a number from 1 to 9999.
+	 * @return The generated name.
+	 */
+	private String generateRandomName() {
```
danthe1st, line 81: `nouns` seems to contain characters with a length until 14 characters, the max length of `adjectives` seems to be 17. Since a random integer with max 4 digits is added, this might exceed the 32-character nickname limit.

The final name should probably be stripped to 32 characters max.

src/main/java/net/javadiscord/javabot/listener/PingableNameListener.java
```diff
@@ -0,0 +1,115 @@
...
+	/**
+	 * Reads strings from the given URL to a {@link List}, removes words with a dash.
+	 * @param url The URL to read.
+	 * @return A {@link List} of Strings.
+	 */
+	private static List<String> readStrings(String url) {
+		List<String> list;
+		try(Scanner scan = new Scanner(new URL(url).openStream()).useDelimiter("\\A")) {
+			String response = scan.next();
+			list = Arrays.stream(response.split("\n")).collect(Collectors.toList());
```
danthe1st, line 97: Why isn't the `Scanner` used for this?

danthe1st (changes requested):

src/main/java/net/javadiscord/javabot/listener/PingableNameListener.java
```diff
@@ -0,0 +1,123 @@
...
+	 */
+	public PingableNameListener() {
+		nouns = readStrings(NOUNS_URL);
+		adjectives = readStrings(ADJECTIVES_URL);
+		log.info("Loaded {} Nouns!", nouns.size());
+		log.info("Loaded {} Adjectives!", adjectives.size());
+	}
+
+	@Override
+	public void onGuildMemberJoin(GuildMemberJoinEvent event) {
```
danthe1st, line 42: The methods `onGuildMemberJoin` and `onGuildMemberUpdateNickname` could be merged together.

src/main/java/net/javadiscord/javabot/listener/PingableNameListener.java
MoonTM-GIT: Yeah, you're right ig

danthe1st (changes requested):

src/main/java/net/javadiscord/javabot/listener/PingableNameListener.java
```diff
@@ -0,0 +1,123 @@
...
+	 * Changes the given {@link Member}s name to a randomly generated one.
+	 * @param member The Member whose name should be changed.
+	 */
+	private void changeName(Member member, TextChannel logChannel) {
+		String newName = generateRandomName();
+		member.modifyNickname(newName).queue();
+		member.getUser().openPrivateChannel()
+				.flatMap(channel -> channel.sendMessageFormat("Your nickname has been set to `%s` since both your user- and nickname's first three characters were deemed as not-pingable.", newName))
+				.queue();
+		logChannel.sendMessageFormat("Changed %s's nickname to %s.", member.getAsMention(), newName).queue();
```
DynxstyGIT, line 65: Use the GuildNotificationService class for this

DynxstyGIT (approved):

src/main/java/net/javadiscord/javabot/listener/PingableNameListener.java
MoonTM-GIT: How would that work?

MoonTM-GIT: What do you mean?

danthe1st:
```java
	@Override
	public void onGuildMemberJoin(GuildMemberJoinEvent event) {
		checkNickname(event.getMember,null);
	}
	@Override
	public void onGuildMemberUpdateNickname(GuildMemberUpdateNicknameEvent event) {
		checkNickname(event.getMember,event.getNewNickname());
	}
	private void checkNickname(Member member, String name){
		  if (!(name==null||isPingable(name)) && !isPingable(member.getUser().getName()) && !canBypassCheck(member)) {
			changeName(member, Bot.config.get(member.getGuild()).getModeration().getLogChannel());
		}
	}
```

src/main/java/net/javadiscord/javabot/listener/PingableNameListener.java
MoonTM-GIT: oh wait, this was done since the bot cannot edit the nicknames of people above itself

danthe1st: What about using `canInteract` instead?

MoonTM-GIT: alright

src/main/java/net/javadiscord/javabot/listener/PingableNameListener.java
danthe1st: I meant that the tokenization of scanner could be used instead of `String#split` (with the delimiter set to `\n`). A `Stream` over all tokens of a `Scanner` can be obtained via [`Scanner#tokens`](https://docs.oracle.com/en/java/javase/18/docs/api/java.base/java/util/Scanner.html).

danthe1st (approved):

PR #282, opened 2022-05-09
Added listener to format "message.txt" source code attachments.

andrewlalis:


src/main/java/net/javadiscord/javabot/listener/CodeAttachmentListener.java
```diff
@@ -0,0 +1,123 @@
...
+public class CodeAttachmentListener extends ListenerAdapter {
+	@Override
+	public void onMessageReceived(@NotNull MessageReceivedEvent event) {
+		if (isCodeMessage(event.getMessage())) {
+			fetchSource(event.getMessage())
+				.thenApply(this::processSource)
+				.thenAcceptAsync(blocks -> {
+					Message lastMsg = event.getMessage();
+					for (var block : blocks) {
+						lastMsg = lastMsg.reply(block).complete();
```
danthe1st, line 30: Add `allowed_mentions` so that this cannot be exploited to ping `@everyone` or similar.

src/main/java/net/javadiscord/javabot/listener/CodeAttachmentListener.java
```diff
@@ -0,0 +1,123 @@
...
+	 * Checks if a message is considered to be a valid "code" message that we
+	 * can process into code blocks. A message is a code message if it contains
+	 * exactly 1 attachment whose name is "message.txt" and whose size does not
+	 * exceed 10KB.
+	 * @param msg The message to check.
+	 * @return True if the message is eligible for processing, or false otherwise.
+	 */
+	private boolean isCodeMessage(Message msg) {
+		return msg.getAttachments().size() == 1 &&
+				msg.getAttachments().get(0).getFileName().equals("message.txt") &&
```
danthe1st, line 46: This should also work with `.java` files.

src/main/java/net/javadiscord/javabot/listener/CodeAttachmentListener.java
```diff
@@ -0,0 +1,123 @@
...
+
+	/**
+	 * Fetches the source code attached to a message.
+	 * @param msg The message to fetch from.
+	 * @return A future that completes when the source code is fetched.
+	 */
+	private CompletableFuture<String> fetchSource(Message msg) {
+		return msg.getAttachments().get(0).retrieveInputStream().thenApply(in -> {
+			try {
+				String s = new String(in.readAllBytes());
```
danthe1st, line 58: Specify the charset.

src/main/java/net/javadiscord/javabot/listener/CodeAttachmentListener.java
```diff
@@ -0,0 +1,123 @@
...
+				in.close();
```
danthe1st, line 59: this should be done either in a `finally`-block or using a `try-with-resources` statement. Such a `try-with-resources` block can be created like this if the variable is declared earlier:
```java
try(InputStream is=in){
	  //code using the InputStream
}
```

src/main/java/net/javadiscord/javabot/listener/CodeAttachmentListener.java
```diff
@@ -0,0 +1,123 @@
...
+ * Listens for messages containing "message.txt" or similar file attachments
+ * containing source code, in order to format the code into one or more code
+ * snippet blocks in the Discord channel instead. The code blocks are sent in
+ * a series of messages that chain their replies to the original message.
+ */
+public class CodeAttachmentListener extends ListenerAdapter {
+	@Override
+	public void onMessageReceived(@NotNull MessageReceivedEvent event) {
+		if (isCodeMessage(event.getMessage())) {
+			fetchSource(event.getMessage())
```
danthe1st, line 25: Handle exceptions thrown by `fetchSource` using `.exceptionally()` or similar.
The `Exception` should be logged using SLF4J.

src/main/java/net/javadiscord/javabot/listener/CodeAttachmentListener.java
```diff
@@ -0,0 +1,123 @@
...
+	/**
+	 * Checks if a message is considered to be a valid "code" message that we
+	 * can process into code blocks. A message is a code message if it contains
+	 * exactly 1 attachment whose name is "message.txt" and whose size does not
+	 * exceed 10KB.
+	 * @param msg The message to check.
+	 * @return True if the message is eligible for processing, or false otherwise.
+	 */
+	private boolean isCodeMessage(Message msg) {
+		return msg.getAttachments().size() == 1 &&
```
danthe1st, line 45: Do we really only want to consider messages with a single attachment and not allow multiple code attachment if the length of all summed up does not exceed 10000 characters (up for discussion)?

src/main/java/net/javadiscord/javabot/listener/CodeAttachmentListener.java
```diff
@@ -0,0 +1,123 @@
...
+
+/**
+ * Listens for messages containing "message.txt" or similar file attachments
+ * containing source code, in order to format the code into one or more code
+ * snippet blocks in the Discord channel instead. The code blocks are sent in
+ * a series of messages that chain their replies to the original message.
+ */
+public class CodeAttachmentListener extends ListenerAdapter {
+	@Override
+	public void onMessageReceived(@NotNull MessageReceivedEvent event) {
```
danthe1st, line 23: It might be better using a command that gets invoked manually (up for discussion).
Not every text attachment should necessarily be automatically converted to a codeblock

src/main/java/net/javadiscord/javabot/listener/CodeAttachmentListener.java
```diff
@@ -0,0 +1,123 @@
...
+		String[] lines = source.split("\\r?\\n");
+		for (int i = 0; i < lines.length; i++) {
+			String line = squishIndentation(lines[i]);
+			// Append the line to the current block, if we can fit it.
+			if (currentBlockLength + line.length() + 1 < 1900) {
+				sb.append(line).append('\n');
+				currentBlockLength += line.length() + 1;
+			} else {
+				// Otherwise, finalize this block and start a new one.
+				blocks.add(String.format("```java%n%s```", sb));
```
danthe1st, line 86: Why are platform-specific line breaks used here even though the message is sent to the discord API instead of saved locally?

src/main/java/net/javadiscord/javabot/listener/CodeAttachmentListener.java
```diff
@@ -0,0 +1,123 @@
...
+				currentBlockLength = line.length() + 1;
+			}
+		}
+		if (currentBlockLength > 0) {
+			blocks.add(String.format("```java%n%s```", sb));
+		}
+		return blocks;
+	}
+
+	private String squishIndentation(String line) {
```
danthe1st, line 101: Why is that needed? As far as I remember, both tabs and spaces should be OK for codeblocks.

src/main/java/net/javadiscord/javabot/listener/CodeAttachmentListener.java
```diff
@@ -0,0 +1,123 @@
...
+	 * snippets for use in discord messages.
+	 * @param source The source to process.
+	 * @return A list of code block strings.
+	 */
+	private List<String> processSource(String source) {
+		List<String> blocks = new ArrayList<>(5);
+		StringBuilder sb = new StringBuilder(2000);
+		int currentBlockLength = 0;
+		String[] lines = source.split("\\r?\\n");
+		for (int i = 0; i < lines.length; i++) {
```
danthe1st, line 78: Why is every line processed separately instead of just sending the message as-is (probably sanitized using JDA's sanitizer)?

danthe1st (changes requested):
Note that this is an implementation of a suggestion that is not yet accepted (<https://canary.discord.com/channels/648956210850299986/973165116226932766>).

src/main/java/net/javadiscord/javabot/listener/CodeAttachmentListener.java
danthe1st, line 30: Aside from that, why did you use a blocking `complete()` call here?

DynxstyGIT: Not necessarily needed, as it is disabled by default

danthe1st: Mentions to `@everyone` are disabled by default but what about role mentions or mass pings of individual users?

DynxstyGIT: Yeah, true

andrewlalis: This way, I am able to chain the message replies such that the first code block replies to the original message, then the next code block replies to the first code block, and then the third code block replies to the second code block, and so on. And I do this using an async block that will be run in whatever executor is being used.

src/main/java/net/javadiscord/javabot/listener/CodeAttachmentListener.java
andrewlalis: Because the content of the source string could be more than 2000 characters, in which case the source should be split at line boundaries.

src/main/java/net/javadiscord/javabot/listener/CodeAttachmentListener.java
andrewlalis: Because 4-space and tab indentation are less readable on smaller screens, which can be especially true when people post larger snippets with heavy indentation.

src/main/java/net/javadiscord/javabot/listener/CodeAttachmentListener.java
andrewlalis: Replaced them with `'\n'`.

src/main/java/net/javadiscord/javabot/listener/CodeAttachmentListener.java
danthe1st: Let's see what the other @Java-Discord/reviewers think on that.

src/main/java/net/javadiscord/javabot/listener/CodeAttachmentListener.java
danthe1st: Note that this would still be possible using `.submit()` instead of `complete()` but I guess it's fine as it happens asynchronously anyways.

andrewlalis: Added a thing to set allowed mentions to an empty set.

src/main/java/net/javadiscord/javabot/listener/CodeAttachmentListener.java
andrewlalis: I made it work with a variety of files:
```java
	private static final Function<Message.Attachment, Boolean> attachmentFilter = attachment ->
			attachment.getFileName().equalsIgnoreCase("message.txt") ||
			Set.of(".java", ".html", ".d", ".c", ".js", ".css").contains(attachment.getFileName().toLowerCase());
```
It can be updated if needed.

src/main/java/net/javadiscord/javabot/listener/CodeAttachmentListener.java
andrewlalis: Done

src/main/java/net/javadiscord/javabot/listener/CodeAttachmentListener.java
andrewlalis: Done

src/main/java/net/javadiscord/javabot/listener/CodeAttachmentListener.java
```diff
@@ -1,34 +1,44 @@
...
  * Listens for messages containing "message.txt" or similar file attachments
  * containing source code, in order to format the code into one or more code
  * snippet blocks in the Discord channel instead. The code blocks are sent in
  * a series of messages that chain their replies to the original message.
  */
+@Slf4j
 public class CodeAttachmentListener extends ListenerAdapter {
+	private static final Function<Message.Attachment, Boolean> attachmentFilter = attachment ->
+			attachment.getFileName().equalsIgnoreCase("message.txt") ||
+			Set.of(".java", ".html", ".d", ".c", ".js", ".css").contains(attachment.getFileName().toLowerCase());
```
danthe1st, line 24: Why would we want to add support for `*.d` and not other file extensions?

src/main/java/net/javadiscord/javabot/listener/CodeAttachmentListener.java
andrewlalis: I changed the code slightly so that it now considers all eligible code attachments.

src/main/java/net/javadiscord/javabot/listener/CodeAttachmentListener.java
danthe1st: Why would we want to add support for `*.d` and not other file extensions?

src/main/java/net/javadiscord/javabot/listener/CodeAttachmentListener.java
andrewlalis: I extracted the main logic for processing attachments of source code, so that it can be made into a command easily.

src/main/java/net/javadiscord/javabot/listener/CodeAttachmentListener.java
andrewlalis: I just put some extensions there for now, but if reviewers want to propose a full list of supported extensions that can be added.

danthe1st: As I said already, using a (message context menu) command that works with every file type may be better as people often don't need this feature and that would allow support for any file but that's up for discussion.

src/main/java/net/javadiscord/javabot/listener/CodeAttachmentListener.java
MoonTM-GIT: I agree with Andrew here

DynxstyGIT (approved):

MoonTM-GIT:
Okay, so are we going to make this a command or keep it the way it is?

DynxstyGIT:
Alright, so, whats up with this

andrewlalis:
You guys have to decide on how you want this feature implemented: either have it automatically detect messages, or add it as a context action thing.

DynxstyGIT:
bump

DynxstyGIT:
What happened to this? I'd say we can use the new preferences system to toggle the automatic file-to-text conversion, and otherwise, have a context commands for this

DynxstyGIT:
Closed due to inactivity - We might pick this up later

PR #287, opened 2022-05-19
Fixed bug where the url parsing just crashes and the suggestion message will be not created.

liquiddevelopmentnet:
Here is the message where the parsing crashed to reproduce this bug:

```
Remove the link (`https://javadiscord.net/`) in Duke 's Activity because it is too long to render and this would be cooler to do this instead.

`Watching https://javadi...` (`Watching https://javadiscord.net/ | 12997 Members`) to
`Watching 12997 Members`

as for the links could be then added to the bot's about me
```

### DIFF
--- a/src/main/java/net/javadiscord/javabot/systems/moderation/AutoMod.java
+++ b/src/main/java/net/javadiscord/javabot/systems/moderation/AutoMod.java
@@ -205,7 +205,8 @@ public class AutoMod extends ListenerAdapter {
 					if (uri.getHost() != null && spamUrls.contains(uri.getHost())) {
 						return true;
 					}
-				} catch (URISyntaxException ignored) {
+				} catch (URISyntaxException e) {
+					log.error("Error while parsing URL: " + url, e);
 				}
 			}
 		}
--- a/src/main/java/net/javadiscord/javabot/systems/moderation/AutoMod.java
+++ b/src/main/java/net/javadiscord/javabot/systems/moderation/AutoMod.java
@@ -202,11 +202,10 @@ public class AutoMod extends ListenerAdapter {
 				String url = urlMatcher.group(0).trim();
 				try {
 					URI uri = new URI(url);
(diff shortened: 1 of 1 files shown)


DynxstyGIT:
Have you tested that?

liquiddevelopmentnet:
Yes

DynxstyGIT (approved):

src/main/java/net/javadiscord/javabot/systems/moderation/AutoMod.java
```diff
@@ -202,11 +202,10 @@ public boolean hasSuspiciousLink(Message message) {
 				String url = urlMatcher.group(0).trim();
 				try {
 					URI uri = new URI(url);
-					if (spamUrls.contains(uri.getHost())) {
+					if (uri.getHost() != null && spamUrls.contains(uri.getHost())) {
 						return true;
 					}
-				} catch (URISyntaxException e) {
-					e.printStackTrace();
+				} catch (URISyntaxException ignored) {
```
danthe1st, line 208: Why did you just ignore the exception without even logging it?

MoonTM-GIT: How does removing the printing of the stack trace stop the problem from occurring in the first place?

liquiddevelopmentnet: As long as it works correctly

danthe1st: Note that `As long as it works correctly` is no excuse for bad code in any way.

MoonTM-GIT: At least put something in the log indicating that this error occurred, swallowing exceptions completely like this is not something you should do

MoonTM-GIT (approved):

PR #297, opened 2022-06-11
Cleaned up ChangeMyMindCommand

### LINKED ISSUE #295: Bug: Changemymind api (nekobot.xyz) is crashing when too many characters are provided
The API our Bot uses for the meme- image generation ([https://nekobot.xyz/](https://docs.nekobot.xyz/)) crashes with `HTTP CODE 501` when the text provided is too long, which results in a empty reply from the bot.

**Steps to reproduce**:
GET https://nekobot.xyz/api/imagegen?type=changemymind&text={Text with 2000 characters}

[File where this should be fixed](https://github.com/Java-Discord/JavaBot/blob/main/src/main/java/net/javadiscord/javabot/systems/commands/ChangeMyMindCommand.java)

andrewlalis:
Fixes #295

### DIFF
--- a/src/main/java/net/javadiscord/javabot/systems/commands/ChangeMyMindCommand.java
+++ b/src/main/java/net/javadiscord/javabot/systems/commands/ChangeMyMindCommand.java
@@ -29,7 +29,7 @@ public class ChangeMyMindCommand implements SlashCommand {
 	 * lower to avoid issues, since the API fails with a 501 error. People
 	 * shouldn't really be putting paragraphs into this anyway.
 	 */
-	private static final int MAX_SEARCH_TERM_LENGTH = 1500;
+	private static final int MAX_SEARCH_TERM_LENGTH = 500;
 
 	@Override
 	public ReplyCallbackAction handleSlashCommandInteraction(SlashCommandInteractionEvent event) {
--- a/src/main/java/net/javadiscord/javabot/systems/commands/ChangeMyMindCommand.java
+++ b/src/main/java/net/javadiscord/javabot/systems/commands/ChangeMyMindCommand.java
@@ -36,7 +36,7 @@ public class ChangeMyMindCommand implements SlashCommand {
 		var hook = event.getHook();
 		String encodedSearchTerm;
 		encodedSearchTerm = URLEncoder.encode(Objects.requireNonNull(event.getOption("text")).getAsString(), StandardCharsets.UTF_8);
-		if (encodedSearchTerm.toCharArray().length > MAX_SEARCH_TERM_LENGTH) {
+		if (encodedSearchTerm.length() > MAX_SEARCH_TERM_LENGTH) {
 			return event.reply("The text you provided is too long. It may not be more than " + MAX_SEARCH_TERM_LENGTH + " characters.");
 		}
 
--- a/src/main/java/net/javadiscord/javabot/systems/commands/ChangeMyMindCommand.java
+++ b/src/main/java/net/javadiscord/javabot/systems/commands/ChangeMyMindCommand.java
@@ -24,14 +24,26 @@ import java.util.Objects;
  */
 @Deprecated
 public class ChangeMyMindCommand implements SlashCommand {
+	/**
+	 * The maximum acceptable length for texts to send to the API. Technically,
+	 * the API supports up to but not including 2000, but we'll make that much
+	 * lower to avoid issues, since the API fails with a 501 error. People
+	 * shouldn't really be putting paragraphs into this anyway.
+	 */
+	private static final int MAX_SEARCH_TERM_LENGTH = 1500;
+
 	@Override
 	public ReplyCallbackAction handleSlashCommandInteraction(SlashCommandInteractionEvent event) {
 		var hook = event.getHook();
-		String encodedSearchTerm = null;
+		String encodedSearchTerm;
 		try {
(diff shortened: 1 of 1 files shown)


DynxstyGIT (approved):

src/main/java/net/javadiscord/javabot/systems/commands/ChangeMyMindCommand.java
```diff
@@ -24,14 +24,26 @@
...
 	@Override
 	public ReplyCallbackAction handleSlashCommandInteraction(SlashCommandInteractionEvent event) {
 		var hook = event.getHook();
-		String encodedSearchTerm = null;
+		String encodedSearchTerm;
 		try {
 			encodedSearchTerm = URLEncoder.encode(Objects.requireNonNull(event.getOption("text")).getAsString(), StandardCharsets.UTF_8.toString());
 		} catch (UnsupportedEncodingException e) {
 			e.printStackTrace();
+			return event.reply("The text you provided is using an unsupported encoding.");
```
danthe1st, line 43: What is that supposed to be? The error occurs when `UTF-8` is unknown.

src/main/java/net/javadiscord/javabot/systems/commands/ChangeMyMindCommand.java
```diff
@@ -24,14 +24,26 @@
  */
 @Deprecated
 public class ChangeMyMindCommand implements SlashCommand {
+	/**
+	 * The maximum acceptable length for texts to send to the API. Technically,
```
danthe1st, line 28: What does `f or text` mean?

src/main/java/net/javadiscord/javabot/systems/commands/ChangeMyMindCommand.java
```diff
@@ -24,14 +24,26 @@
...
+	 * the API supports up to but not including 2000, but we'll make that much
+	 * lower to avoid issues, since the API fails with a 501 error. People
+	 * shouldn't really be putting paragraphs into this anyway.
```
danthe1st, line 31: Are you sure that 1500 characters is the correct length? What about unicode letters with multiple codepoints?

andrewlalis: There is no documented length. So I don't know, but the issue said that 2000 characters breaks it. So limiting much lower is just an attempt at making it work in all cases.

src/main/java/net/javadiscord/javabot/systems/commands/ChangeMyMindCommand.java
andrewlalis: You're reading it wrong. It says "The maximum acceptable length for texts to send to the API"

src/main/java/net/javadiscord/javabot/systems/commands/ChangeMyMindCommand.java
andrewlalis: Removed it as `URLEncoder#encode` doesn't even throw that when given a Charset instead of a String.

src/main/java/net/javadiscord/javabot/systems/commands/ChangeMyMindCommand.java
danthe1st: My fault. My device cut that off.

src/main/java/net/javadiscord/javabot/systems/commands/ChangeMyMindCommand.java
danthe1st: Have you tested whether ot works with 1500 characters?

andrewlalis: Yeah apparently 1500 is still too much, so I changed to 500 which does work.

danthe1st: Do unicode characters with multiple codepoints (like emojis) change anything?

andrewlalis: Well, I did this test:
```java
class Scratch {
	public static void main(String[] args) {
		String text = "\uD83D\uDE00".repeat(250);
		String encodedSearchTerm;
		encodedSearchTerm = URLEncoder.encode(text, StandardCharsets.UTF_8);
		System.out.println(encodedSearchTerm.length());
		System.out.println(encodedSearchTerm.toCharArray().length);
		System.out.println(encodedSearchTerm.getBytes(StandardCharsets.UTF_8).length);
	}
}
```
Which shows that my logic works correctly to determine the actual length of a string with many multiple codepoint characters, where it outputs `3000` as length for all cases.

So no, emojis don't change anything; it still works.

danthe1st: Can you try a binary search or similar for finding out tze maximum length?

andrewlalis: Well, I'm pretty sure it's just a limitation due to the length of the query string. But it's pointless to try and find an exact max since with 500 characters the text already goes off the image.

DynxstyGIT (approved):

PR #300, opened 2022-06-11
Improved /help account Formatting

andrewlalis:
We've gotten several complaints about this:
https://discord.com/channels/648956210850299986/752535909228085348/984352629884862496

So I have improved the UI to:
- Show the current level first.
- Then show the progress to the next level
- Removed decimals from the XP since it takes up space and nobody needs it to be that precise.

### DIFF
--- a/src/main/java/net/javadiscord/javabot/systems/help/commands/subcommands/HelpAccountSubcommand.java
+++ b/src/main/java/net/javadiscord/javabot/systems/help/commands/subcommands/HelpAccountSubcommand.java
@@ -73,16 +73,23 @@ public class HelpAccountSubcommand implements SlashCommand {
 	}
 
 	private String formatExperience(Guild guild, HelpAccount account) {
-		double current = account.getExperience() - account.getLastExperienceGoal(guild);
-		Pair<Role, Double> role = account.getNextExperienceGoal(guild);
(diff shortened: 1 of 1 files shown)


MoonTM-GIT (approved):

DynxstyGIT:
looks good!
we thought of changing the progress bar itself too though, as it’s on 2 separate lines and looks ugly on iOS 

image

Any ideas on that?

andrewlalis:
@DynxstyGIT you can use block elements: https://en.wikipedia.org/wiki/Block_Elements

This might be a bit trickier, but it can be used to maybe make something that looks a little nicer.

MoonTM-GIT:
I think we ready had some pretty good ideas in the staff chat, I can't find it right now but I know they were from Liquid

liquiddevelopmentnet:
image
We had this one here.

liquiddevelopmentnet:
Also, I would suggest to include the total help XP in `/help account`
ref: https://canary.discord.com/channels/648956210850299986/752535909228085348/985221082791419934

andrewlalis:
I think I'll just keep the current progress bar for this pull request, and make a separate pull request for an updated progress bar, that way I can add tests without cluttering things up. But I did add a total XP thing so this should be ready to go.

DynxstyGIT (approved):

PR #301, opened 2022-06-11
Improved Javadoc For Db Interfaces, Added countf Function

andrewlalis:


src/main/java/net/javadiscord/javabot/data/h2db/DbActions.java
```diff
@@ -118,6 +124,23 @@ public static long count(String query) {
...
+	 * <p>
+	 *     <strong>WARNING</strong>: This method should NEVER be called with
+	 *     user-provided data.
+	 * </p>
+	 *
+	 * @param queryFormat The format string.
+	 * @param args The set of arguments to pass to the formatter.
+	 * @return The count.
+	 */
+	public static long countf(String queryFormat, Object... args) {
```
danthe1st, line 140: What's the purpose of that method? Why not use `PreparedStatement` for this?

andrewlalis: It makes count operations more compact and removes them from the specificity of jdbc.
```java
countf("SELECT COUNT(id) FROM entries WHERE id = %d", 4);
```
versus
```java
count("SELECT COUNT(id) FROM entries WHERE id = ?", s -> s.setLong(1, 4));
```

danthe1st: I don't think such a method should be created since it may be used accidentally for user input without reading the documentation.

PR #307, opened 2022-06-13
Add a minimum server membership time to send media. (fix #305)

### LINKED ISSUE #305: Add a minimum server membership time to send media.
Since there were many raids lately with multiple accounts sending Images, GIF's & Videos of NSFW content, i would recomment to create a minimum time (configurable per server config) that the user must have been on the server to send media. Of course, this wouldn't really solve the problem but its a good solution since we can't really do something else.

liquiddevelopmentnet:
A video of the point of view from the user:
https://user-images.githubusercontent.com/76592300/173371233-b06e1477-2448-4d05-a6b5-379e74e167b3.mp4

The log channel:
DiscordCanary_aSUA4dezpA

A screenshot of the new guild-configuration entry:
idea64_eGxc8eZhSJ

See Issue #305 for more details.

### DIFF
--- a/src/main/java/net/javadiscord/javabot/data/config/guild/ModerationConfig.java
+++ b/src/main/java/net/javadiscord/javabot/data/config/guild/ModerationConfig.java
@@ -99,5 +99,7 @@ public class ModerationConfig extends GuildConfigItem {
 		return this.getGuild().getRoleById(this.expertRoleId);
 	}
 
-	public long getMinActiveHoursToSendMedia() { return this.minActiveHoursToSendMedia; }
+	public long getMinActiveHoursToSendMedia() {
+		return this.minActiveHoursToSendMedia;
+	}
 }
--- a/src/main/java/net/javadiscord/javabot/data/config/guild/ModerationConfig.java
+++ b/src/main/java/net/javadiscord/javabot/data/config/guild/ModerationConfig.java
@@ -21,6 +21,8 @@ public class ModerationConfig extends GuildConfigItem {
 	private long adminRoleId;
 	private long expertRoleId;
 
+	private long minActiveHoursToSendMedia;
+
 	/**
 	 * ID of the share-knowledge channel.
 	 */
@@ -96,4 +98,6 @@ public class ModerationConfig extends GuildConfigItem {
 	public Role getExpertRole() {
 		return this.getGuild().getRoleById(this.expertRoleId);
 	}
+
+	public long getMinActiveHoursToSendMedia() { return this.minActiveHoursToSendMedia; }
 }
--- a/src/main/java/net/javadiscord/javabot/systems/moderation/AutoMod.java
+++ b/src/main/java/net/javadiscord/javabot/systems/moderation/AutoMod.java
@@ -8,9 +8,7 @@ import net.dv8tion.jda.api.entities.Message.Attachment;
 import net.dv8tion.jda.api.events.message.MessageReceivedEvent;
 import net.dv8tion.jda.api.events.message.MessageUpdateEvent;
 import net.dv8tion.jda.api.hooks.ListenerAdapter;
-import net.dv8tion.jda.api.requests.restaction.MessageAction;
 import net.javadiscord.javabot.Bot;
-import net.javadiscord.javabot.data.h2db.message_cache.model.CachedMessage;
 import net.javadiscord.javabot.systems.commands.IdCalculatorCommand;
 import net.javadiscord.javabot.systems.moderation.warn.model.WarnSeverity;
 import net.javadiscord.javabot.systems.notification.GuildNotificationService;
@@ -124,7 +122,7 @@ public class AutoMod extends ListenerAdapter {
 		if (Bot.config.get(message.getGuild()).getModeration().getMinActiveHoursToSendMedia() > 0 && message.getMember() != null) {
 			List<String> imageOrVideo = new ArrayList<>();
 			for (Attachment attachment : message.getAttachments()) {
-				if ((attachment.isImage() || attachment.isVideo())) {
+				if (attachment.isImage() || attachment.isVideo()) {
 					imageOrVideo.add(attachment.getUrl());
 				}
 			}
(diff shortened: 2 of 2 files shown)


src/main/java/net/javadiscord/javabot/data/config/guild/ModerationConfig.java
```diff
@@ -21,6 +21,8 @@ public class ModerationConfig extends GuildConfigItem {
 	private long adminRoleId;
 	private long expertRoleId;
 
+	private long minActiveHoursToSendMedia;
```
MoonTM-GIT, line 24: Remove this empy line
```suggestion
	private long expertRoleId;
	private long minActiveHoursToSendMedia;
```

src/main/java/net/javadiscord/javabot/data/config/guild/ModerationConfig.java
```diff
@@ -96,4 +98,6 @@ public Role getAdminRole() {
 	public Role getExpertRole() {
 		return this.getGuild().getRoleById(this.expertRoleId);
 	}
+
+	public long getMinActiveHoursToSendMedia() { return this.minActiveHoursToSendMedia; }
```
MoonTM-GIT: Remove this, the getters for Roles/Channels are there to return them as an Role or Channel object instead of the id

src/main/java/net/javadiscord/javabot/data/config/guild/ModerationConfig.java
```diff
@@ -96,4 +98,8 @@ public Role getAdminRole() {
 	public Role getExpertRole() {
 		return this.getGuild().getRoleById(this.expertRoleId);
 	}
+
+	public long getMinActiveHoursToSendMedia() {
```
MoonTM-GIT, line 102: Remove this, we have made custom getters for Channels & Roles to return their respective objects instead of plain IDs. This is already done by the `@Data` annotation

src/main/java/net/javadiscord/javabot/systems/moderation/AutoMod.java
```diff
@@ -85,6 +90,8 @@ private boolean canBypassAutomod(Member member) {
 	 * @param message the {@link Message} that should be checked
 	 */
 	private void checkNewMessageAutomod(@Nonnull Message message) {
+		AtomicBoolean consideredSpam = new AtomicBoolean(false);
```
MoonTM-GIT, line 93: Why is this needed? (also applies to the usage at the bottom)

src/main/java/net/javadiscord/javabot/systems/moderation/AutoMod.java
```diff
@@ -228,6 +266,47 @@ public boolean hasAdvertisingLink(Message message) {
 		return false;
 	}
 
+	/**
+	 * Builds an embed that tells the user that he/she has to wait a certain amount of time before posting media.
```
MoonTM-GIT, line 270:
```suggestion
	 * Builds an embed, that tells the user that they have to wait a certain amount of time before posting media.
```

src/main/java/net/javadiscord/javabot/systems/moderation/AutoMod.java
```diff
@@ -106,9 +114,39 @@ private void checkNewMessageAutomod(@Nonnull Message message) {
 					.filter(msg -> (message.getTimeCreated().toEpochSecond() - msg.getTimeCreated().toEpochSecond()) < 6).count();
 			if (spamCount > 5) {
 				handleSpam(message, message.getMember());
+				consideredSpam.set(true);
 			}
 		});
 
+		// check if the user has the required membership time to send media
```
MoonTM-GIT, line 121:
```suggestion
		// Checks, whether the user has been on the server long enough to send media.
```

MoonTM-GIT (changes requested):

src/main/java/net/javadiscord/javabot/systems/moderation/AutoMod.java
liquiddevelopmentnet: Because of the use in the lambda

liquiddevelopmentnet: Because its not that useful if the bot spams the embed for not enough time and spam, since this would be spam again. My idea was that when it was already considered as spam its not necessary to send the other embed too.

danthe1st (changes requested):
Sending images right after joining is an integral part of the JavaDiscord's functionality and required to still function.
For example, many people come to the server in order to ask questions. They often send images as a part of their question (e.g. for demonstrating functionality).
Hence we cannot deny sending images for new users.

liquiddevelopmentnet:
> Sending images right after joining is an integral part of the JavaDiscord's functionality and required to still function. For example, many people come to the server in order to ask questions. They often send images as a part of their question (e.g. for demonstrating functionality). Hence we cannot deny sending images for new users.

They could just send text. What is your idea to prevent the things to happen then?

MoonTM-GIT:
I agree with dan here, this is quite important.

liquiddevelopmentnet:
What is your idea to solve the problem then?

MoonTM-GIT:
Ban the people that do it, there's not much we can do if Discord's filter doesn't work properly. Dan is right, being able to post pictures is integral to asking questions sometimes.

liquiddevelopmentnet:
I see. Next time say something before I write the code

MoonTM-GIT:
Sorry, but you can't expect someone to respond to an issue within 2.5 hours on a weekday. I have other responsibilities besides administrating this Discord.

danthe1st:
> They could just send text. 
They cannot always do that. In many cases they could (e.g. with images of code, tasks, etc.) but often (especially with graphical applications), images are necessary for explaining problems.

Furthermore, I think that it would be possible to bypass your restriction by editing messages or using links to images/videos.

liquiddevelopmentnet:
> > They could just send text.
> > They cannot always do that. In many cases they could (e.g. with images of code, tasks, etc.) but often (especially with graphical applications), images are necessary for explaining problems.
> 
> Furthermore, I think that it would be possible to bypass your restriction by editing messages or using links to images/videos.

Okay, but that could be fixed

MoonTM-GIT:
How are you going to fix this?
> They cannot always do that. In many cases they could (e.g. with images of code, tasks, etc.) but often (especially with graphical applications), images are necessary for explaining problems.

As I said, we can't do anything except maybe trying to get more mods from more time-zones.

liquiddevelopmentnet:
Then, why are the applications closed?

MoonTM-GIT:
It was a solution I suggested, I didn't say that's what we are actively trying right now.

danthe1st:
Because we normally have sufficient moderators. This is mainly a temporary issue of multiple people not having time.

liquiddevelopmentnet:
This is mainly a issue of people sleeping when they are doing it.

danthe1st:
Deleting their message would lead to frustration of users if they see `you cannot post images because you are too new` instead of informing them about the help channels. That would cause many people to leave right after joining.

PR #308, opened 2022-06-16
replace f*cks with hugs

danthe1st:
https://lkml.org/lkml/2018/12/1/105

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -36,6 +36,9 @@ dependencies {
 
     // Quartz scheduler
     implementation("org.quartz-scheduler:quartz:2.3.2")
+    
+    // Webhooks
+    implementation("club.minnced:discord-webhooks:0.8.0")
 
     // Lombok Annotations
     compileOnly("org.projectlombok:lombok:1.18.24")
--- a/src/main/java/net/javadiscord/javabot/Bot.java
+++ b/src/main/java/net/javadiscord/javabot/Bot.java
@@ -142,7 +142,8 @@ public class Bot {
 				new HelpChannelListener(),
 				new ShareKnowledgeVoteListener(),
 				new JobChannelVoteListener(),
-				new PingableNameListener()
+				new PingableNameListener(),
+				new HugListener()
 		);
 	}
 }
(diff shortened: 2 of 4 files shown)


DynxstyGIT (approved):

danthe1st:
Note that the username/avatar cannot be set as JDA5 currently doesn't have webhook support.

DynxstyGIT:
you might want to take a look at this: https://github.com/MinnDevelopment/discord-webhooks

DynxstyGIT (approved):

DynxstyGIT:
It would be great if you (or someone else) could test this functionality tho

danthe1st:
I tested it with normal text, images, images devlared as spoilers and in a thread. However, i still think someone else should test it in isolation.

Note that obtaining non-image attachments currently fail due to a bug (fixed but the fix is not yet published) in JDA.

MoonTM-GIT:
I'll test it a bit

MoonTM-GIT:
Okay, so something that is kind of expected but I would still like to mention, is that voice-text channels are not supported by JDA yet (https://github.com/DV8FromTheWorld/JDA/pull/2072), so it doesn't work in these obviously

Other than that I haven't found anything that didn't work as expected, except for non-image attachments which you already mentioned

MoonTM-GIT (approved):
Not tested

MoonTM-GIT:
Not sure if we still want someone else to test this though

DynxstyGIT:
would be cool if you cloud extract this webhook stuff into a separate method/class tho

DynxstyGIT:
(for things like the messagelink listener)

MoonTM-GIT:
Oh, something I just noticed, if I ping someone in my initial message the ping will (as it shouldn't) not go through to the message the bot sends. This might lead to confusion as someone might be looking for the message he was pinged in, which he can't find that easily since it doesn't have that orange mark, so we might want to disable this functionality for messages that include any pings

danthe1st:
Yes but if that feature would accept pings, it should be capped to e.g. a maximum of three mentions per message.
Aside from that, I might need to explicitely exclude the suggestion channel.

MoonTM-GIT:
I wasn't suggesting to let pings through, that would lead to more confusion as people would he shown two pings, instead it could just not replace anything in messages that include pings to cause no confusion about pings

danthe1st:
> would be cool if you cloud extract this webhook stuff into a separate method/class tho

Done

> I wasn't suggesting to let pings through, that would lead to more confusion as people would he shown two pings, instead it could just not replace anything in messages that include pings to cause no confusion about pings

Done

DynxstyGIT (approved):

danthe1st:
The functionality is implemented and has been tested.

MoonTM-GIT:
yay

DynxstyGIT:
this is a huge step for the java community

PR #318, opened 2022-07-23
Move Conversation

### LINKED ISSUE #259: Add a `/move-conversation` feature, inspired by the Rust Community Discord
I guess this could be quite useful

DynxstyGIT:
Closes #259 

This PR introduces `/move-conversation`, which makes it easier for users to easily move conversations to other channels by clicking the links the bot provides. It is highly inspired by the Rust Community Discord, where this works quite well. 

image
image

### DIFF
--- a/src/main/java/net/javadiscord/javabot/systems/user_commands/MoveConversationCommand.java
+++ b/src/main/java/net/javadiscord/javabot/systems/user_commands/MoveConversationCommand.java
@@ -60,9 +60,9 @@ public class MoveConversationCommand extends SlashCommand {
 			return;
 		}
 		event.deferReply(true).queue();
-		sendMoveToChannelMessage(event, channel).queue(movedFrom ->
-				sendMovedFromChannelMessage(event, channel, movedFrom).queue(movedTo ->
-						editMoveToChannelMessage(event, movedFrom, movedTo).queue(s ->
+		sendMovedFromChannelMessage(event, channel).queue(movedTo ->
+				sendMoveToChannelMessage(event, channel, movedTo).queue(movedFrom ->
+						editMovedFromChannelMessage(event, movedFrom, movedTo).queue(s ->
 								event.getHook().sendMessage("Done!").queue()
 						)
 				)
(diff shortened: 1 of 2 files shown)


danthe1st (changes requested):
Who is allowed to move channels? There neither seem to be any permission check (staff and maybe certified helpers if the conversation is moved to a help channel) nor a voting system (not sure whether a voting system is a good idea).

Aside from that, where is the command descriptor?

MoonTM-GIT:
I feel like we dont need to restrict this command, but I would change the text to make it clear it's a suggestion (for example just _suggested__ instead of _requested_)

danthe1st:
If it's just about suggesting to move a conversation, what's the point of having a command for that over just mentioning it in the chat?
Furthermore, if this is kept as-is, the command name should reflect that.

MoonTM-GIT:
> If it's just about suggesting to move a conversation, what's the point of having a command for that over just mentioning it in the chat?

To make it easier, just like tags/custom commands

MoonTM-GIT:
And making clear it's a suggestion could just be done in the command description

danthe1st:
Most people don't read the command description. The title `/move-conversation` strongly suggests that the command is actually about moving the conversation, not about suggesting to do so.

MoonTM-GIT:
It's obvious nobody has the ability to force someone to go somewhere else, and even if they do think that they'll see once they have used the command as the text clearly says **suggested**

danthe1st:
I am talking about people that use the command (also for the first time), not other people seeing tge message. The name `/move-conversation` looks like it is a restricted command discouraging its use for normal usets.

DynxstyGIT:
I think the name is quite fitting.

src/main/java/net/javadiscord/javabot/systems/user_commands/MoveConversationCommand.java
```diff
@@ -0,0 +1,107 @@
...
+			return;
+		}
+		if (isInvalidChannel(event.getMember(), channel)) {
+			Responses.warning(event, "Invalid Channel", "You're not allowed to move the conversation to %s", channel.getAsMention()).queue();
+			return;
+		}
+		event.deferReply(true).queue();
+		sendMovedFromChannelMessage(event, channel).queue(movedTo ->
+				sendMoveToChannelMessage(event, channel, movedTo).queue(movedFrom ->
+						editMovedFromChannelMessage(event, movedFrom, movedTo).queue(s ->
```
danthe1st, line 65: Qe should probably also add a case for when sending the messages fails.

DynxstyGIT: Done

liquiddevelopmentnet (approved):

DynxstyGIT:
thanks liquid

liquiddevelopmentnet:
> thanks liquid

You're welcome json

DynxstyGIT:
@danthe1st can this be merged?

MoonTM-GIT (approved):

danthe1st (approved):

PR #330, opened 2022-07-26
Improved Redeploy System

MoonTM-GIT:
This PR aims to improve the redeploy system a bit. This deprecates the method described in #195.
This now requires a `sh` to be present at `/bin/sh`, in addition to an updated redeploy script to support all new features.

This is the redeploy script we now use to re-compile the Bot:
```sh
LOCATION="Location" #e.g /home/myUser/JavaBot

cd $LOCATION
rm -rf $LOCATION/redeploy/

git clone https://github.com/Java-Discord/JavaBot/ redeploy
cd redeploy

bash gradlew :shadow

if [ $? -eq 0 ]; then
        rm $LOCATION/JavaBot.jar
        mv $LOCATION/redeploy/build/libs/JavaBot-1.0.0-SNAPSHOT-all.jar $LOCATION/JavaBot.jar
else
        echo COMPILATION FAILED
        echo Did not replace file.
fi
```

As you can see this now only replaces the `.jar` if the compilation was successful. This is also used in the Bot to cancel the redeploy if the compilation did not succeed.

The location of this script has to be set in the Systems config. E.g. `"redeployScriptLocation": "/home/myUser/JavaBot/redeploy.sh",`

We still use a `systemd` service to run the bot, but we don't run this script before every start anymore, instead we only use this script for the `/redeploy` command. This allows us to restart the bot without waiting for it to compile again if we don't want that.

### DIFF
--- a/src/main/java/net/javadiscord/javabot/systems/staff_commands/RedeployCommand.java
+++ b/src/main/java/net/javadiscord/javabot/systems/staff_commands/RedeployCommand.java
@@ -58,6 +58,7 @@ public class RedeployCommand extends SlashCommand {
 			ExceptionLogger.capture(e, getClass().getSimpleName());
 		}
 		Bot.getMessageCache().synchronize();
+		event.getJDA().shutdown();
 		System.exit(0);
 	}
 }
--- a/src/main/java/net/javadiscord/javabot/systems/staff_commands/RedeployCommand.java
+++ b/src/main/java/net/javadiscord/javabot/systems/staff_commands/RedeployCommand.java
@@ -2,6 +2,8 @@ package net.javadiscord.javabot.systems.staff_commands;
 
 import com.dynxsty.dih4jda.interactions.commands.SlashCommand;
 import lombok.extern.slf4j.Slf4j;
+import net.dv8tion.jda.api.EmbedBuilder;
+import net.dv8tion.jda.api.entities.MessageEmbed;
 import net.dv8tion.jda.api.events.interaction.command.SlashCommandInteractionEvent;
 import net.dv8tion.jda.api.interactions.commands.DefaultMemberPermissions;
 import net.dv8tion.jda.api.interactions.commands.build.Commands;
@@ -20,7 +22,7 @@ import java.io.IOException;
  * This only works if the way the bot is hosted is set up correctly with a bash script that handles
  * compilation and a service set up that automatically restarts the Bot..
  * <p>
- * I have explained how we do it in https://github.com/Java-Discord/JavaBot/pull/195
+ * I have explained how we do it in <a href="https://github.com/Java-Discord/JavaBot/pull/330">PR #330</a>.
  */
 @Slf4j
 public class RedeployCommand extends SlashCommand {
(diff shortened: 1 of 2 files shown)


danthe1st:
It might be worthy to use a variable in the redeploy script for spexifyibg the bot directory (`/home/$USR/JavaBot/`).

MoonTM-GIT:
> Aside from that, it would be possible to build the JAR with GitHub Actions and let the bot query the GitHub API and download that.

I know, I also thought about that but I think I personally prefer to compile it myself.

> Furthermore, it is possible to update the bot purely in Java using ClassLoaders (requires a custom starter though).

What would be the benefit of that? This script is easily executable if we don't have anything else running and it seems pretty simple.

danthe1st:
> What would be the benefit of that? 

It would be platform independent, not require the script to be located at a specific path and doesn't require any external scripts at all.

MoonTM-GIT:
> It would be platform independent, not require the script to be located at a specific path and doesn't require any external scripts at all.

Yeah, but it also makes it harder to operate from outside the command which is quite important in my opinion.

DynxstyGIT (approved):

src/main/java/net/javadiscord/javabot/systems/staff_commands/RedeployCommand.java
```diff
@@ -39,8 +42,23 @@ public void execute(@NotNull SlashCommandInteractionEvent event) {
...
+				return;
+			} else {
+				event.getHook().sendMessage("Compilation successful, restarting...").queue();
+			}
+		} catch (InterruptedException | IOException e) {
+			ExceptionLogger.capture(e, getClass().getSimpleName());
+		}
 		Bot.getMessageCache().synchronize();
+		event.getJDA().shutdown();
 		System.exit(0);
```
danthe1st, line 62: We may want to move this to its own listener so that it is called when shutting down JDA completed.

src/main/java/net/javadiscord/javabot/systems/staff_commands/RedeployCommand.java
```diff
@@ -39,8 +42,23 @@ public void execute(@NotNull SlashCommandInteractionEvent event) {
 			return;
 		}
 		log.warn("Redeploying... Requested by: " + event.getUser().getAsTag());
-		event.reply("**Redeploying...** This may take some time.").queue();
+		event.reply("Redeploying, this may take a few minutes.").queue();
+		try {
+			Process p = new ProcessBuilder("/bin/sh", Bot.getConfig().getSystems().getRedeployScriptLocation()).start();
+			p.waitFor();
+			String result = new String(p.getInputStream().readAllBytes());
+			if (result.contains("COMPILATION FAILED")) {
```
danthe1st, line 50: It might be better to use the exit code of the script for checking whether it was successful or not.
For example, there could be errors other than the Gradle script failing (e.g. with `git clone`).

For failing with any error, `set -e` could be used in the script in order to abort the script as soon as any command fails.

danthe1st (changes requested):

src/main/java/net/javadiscord/javabot/systems/staff_commands/RedeployCommand.java
MoonTM-GIT: Yup, will do that soon™️

DynxstyGIT:
Closed due to inactivity

PR #340, opened 2022-08-01
Refactored NotificationService

DynxstyGIT:


src/main/java/net/javadiscord/javabot/systems/notification/GuildNotificationService.java
```diff
@@ -1,62 +1,50 @@
...
+import java.util.function.Function;
 
 /**
- * Sends notifications within a single {@link Guild}.
+ * Handles all sorts of guild notifications.
  */
 @Slf4j
-public final class GuildNotificationService extends NotificationService {
+@RequiredArgsConstructor(access = AccessLevel.PACKAGE)
+public final class GuildNotificationService extends NotificationService.MessageChannelNotification {
```
danthe1st, line 19: Maybe make this class package private?

src/main/java/net/javadiscord/javabot/systems/notification/NotificationService.java
```diff
@@ -1,35 +1,59 @@
...
+
+	public static @NotNull QOTWNotificationService withQOTW(Guild guild, User user) {
+		return new QOTWNotificationService(user, guild);
+	}
+
+	/**
+	 * Abstract class which streamlines the logic of sending messages to a {@link MessageChannel}.
+	 */
+	@Slf4j
+	protected abstract static class MessageChannelNotification {
```
danthe1st, line 43: Why is this class `protected` if no subclasses of `NotificationService ` are allowed?
Wouldn't package private be sufficient?

src/main/java/net/javadiscord/javabot/systems/notification/NotificationService.java
```diff
@@ -1,35 +1,59 @@
...
+ * Handles all types of guild & user notifications.
  */
-@Slf4j
-public abstract class NotificationService {
-	void sendDirectMessageNotification(User user, MessageEmbed message) {
-		user.openPrivateChannel().queue(
-				channel -> sendMessageChannelNotification(channel, message),
-				error -> log.warn("Could not send private Notification to User " + user.getAsTag())
-		);
+public class NotificationService {
```
danthe1st, line 17: Make this class `final` to document the fact that is is not possible to create subclasses of `NotificationService` (due to the private constructor)

danthe1st (changes requested):

src/main/java/net/javadiscord/javabot/systems/notification/GuildNotificationService.java
DynxstyGIT: Why? This wouldn't let me access the classes' methods from the outside

danthe1st: I forgot to delete that comment after looking at the other classes.

danthe1st (approved):

PR #346, opened 2022-08-04
QOTW fixes and making sure RestActions are queued

danthe1st:
This PR adds `.queue` to all `RestAction`s missing it and adds `@CheckReturnValue` to the methods in the `Responses` class so that static code analysis tools can detect that mistake.

Furthermore, it includes the following fixes to `/qotw-view`:
- Answer threads are automatically locked when created by the `/qotw-view` command.
- The order of messages in `/qotw-view` answer threads is fixed.
- The column `text` was renamed to `TEXT` in the `QOTW_QUESTION` (and the corresponding query) in order to fix an exception in `/qotw-view`.
- A webhook is used for sending QOTW answer messages (both for best answers and the `/qotw-view` command) instead of sending the messages directly.

### DIFF
--- a/src/main/java/net/javadiscord/javabot/systems/qotw/commands/view/QOTWViewAnswerSubcommand.java
+++ b/src/main/java/net/javadiscord/javabot/systems/qotw/commands/view/QOTWViewAnswerSubcommand.java
@@ -73,8 +73,7 @@ public class QOTWViewAnswerSubcommand extends SlashCommand.Subcommand {
 													MessageActionUtils.copyMessagesToNewThread(event.getGuildChannel().asStandardGuildMessageChannel(),
 															buildQOTWInfoEmbed(submission, event.getMember() == null ? event.getUser().getName() : event.getMember().getEffectiveName()),
 															"QOTW #" + submission.getQuestionNumber(),
-															history
-																.getRetrievedHistory()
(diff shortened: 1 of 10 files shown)


DynxstyGIT (approved):

src/main/java/net/javadiscord/javabot/systems/qotw/commands/view/QOTWViewAnswerSubcommand.java
```diff
@@ -63,8 +73,16 @@ public void execute(@NotNull SlashCommandInteractionEvent event) {
 													MessageActionUtils.copyMessagesToNewThread(event.getGuildChannel().asStandardGuildMessageChannel(),
 															buildQOTWInfoEmbed(submission, event.getMember() == null ? event.getUser().getName() : event.getMember().getEffectiveName()),
 															"QOTW #" + submission.getQuestionNumber(),
-															history.getRetrievedHistory(),
-															() -> Responses.success(event.getHook(), "View Answer", "Answer copied successfully"))),
+															history
+																.getRetrievedHistory()
```
DynxstyGIT, line 77:
```suggestion
															history.getRetrievedHistory()
```

DynxstyGIT (approved):

PR #370, opened 2022-12-01
use damerau-levenshtein distance instead of direct comparison

gsayson:
In the original implementation of `HugListener`, a simple equality check was used. This resulted in plenty of easy bypasses such as "fug" and "fck".

This PR improves the detection mechanism such that instead of an *equality* check being used, a *similarity* check is used to identify some bypasses through the Damerau-Levenshtein algorithm. 

Given a string, the algorithm measures the number of transpositions, substitutions, insertions, or deletions, of a character in order to match the string `"fuck"`.

### DIFF
--- a/src/main/java/net/javadiscord/javabot/listener/HugListener.java
+++ b/src/main/java/net/javadiscord/javabot/listener/HugListener.java
@@ -11,6 +11,7 @@ import net.dv8tion.jda.api.hooks.ListenerAdapter;
 import net.javadiscord.javabot.data.config.BotConfig;
 import net.javadiscord.javabot.systems.moderation.AutoMod;
 import net.javadiscord.javabot.util.WebhookUtil;
+import org.jetbrains.annotations.NotNull;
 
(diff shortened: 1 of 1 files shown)


DynxstyGIT:
Hey, this is actually pretty cool, but I don't know if we really want something like this as we actually don't want to censor "fuck" by itself. The current implementation is more like a joke, which one can bypass easily

If we decide to do this, this behaviour should be controllable using the /preferences command imo

danthe1st:
Note that for _censoring_, reading/reacting to automod logs would be more effective (if we ever plan to do that).

DynxstyGIT:
Yeah, but we don't want to censor stuff. And we never will

DynxstyGIT:
The only thing we ever did was "Glyph", "Deveden" and "Coda", but that was just for fun to annoy big rat :^)

liquiddevelopmentnet:
Pull request of the year

DynxstyGIT:
Closed due to inactivity

PR #377, opened 2022-12-17
QOTW leaderboard improvements

danthe1st:
This PR makes sure that unranked users do not show up in the leaderboard:
image
It also changes the leaderboard and other statistics to only show points from the current month.
Aside from that, it adds and removes the OQTW Champion role automatically at the beginning of each month.


DynxstyGIT (approved):

src/main/java/net/javadiscord/javabot/systems/qotw/QOTWPointsService.java
```diff
@@ -121,15 +124,34 @@ public List<QOTWAccount> getTopAccounts(int amount, int page) {
...
+							member.getGuild().addRoleToMember(member, qotwChampionRole).queue();
+							for (Member m : member.getGuild().getMembersWithRoles(qotwChampionRole)) {
+								if (getOrCreateAccount(m.getIdLong()).getPoints() < best.getPoints()) {
+									m.getGuild().removeRoleFromMember(m, qotwChampionRole).queue();
+								}
+							}
+						}
+					});
+			}
+			return newScore;
```
DynxstyGIT, line 154: What about scheduling a job that does this at the end of each month? It's the "QOTW Champion" of the last month then

DynxstyGIT (changes requested):

src/main/java/net/javadiscord/javabot/systems/qotw/QOTWPointsService.java
danthe1st: Done

PR #401, opened 2023-03-26
timeout users when exceeding specific warn severity

danthe1st:
This PR automatically timeouts users when they reach a specific severity (50 by default) for a specific amount of time (2h by default).

### DIFF
--- a/src/main/java/net/javadiscord/javabot/data/config/guild/ModerationConfig.java
+++ b/src/main/java/net/javadiscord/javabot/data/config/guild/ModerationConfig.java
@@ -62,7 +62,7 @@ public class ModerationConfig extends GuildConfigItem {
 	private int timeoutSeverity = 50;
 
 	/**
-	 * The duration to timeout users when they exceeded {@link #timeoutSeverity}.
+	 * The duration (in hours) to timeout users when they exceeded {@link #timeoutSeverity}.
 	 */
 	private int warnTimeoutHours = 2;
 
--- a/src/main/java/net/javadiscord/javabot/data/config/guild/ModerationConfig.java
+++ b/src/main/java/net/javadiscord/javabot/data/config/guild/ModerationConfig.java
@@ -56,6 +56,17 @@ public class ModerationConfig extends GuildConfigItem {
 	private int maxWarnSeverity = 100;
 
 	/**
+	 * The maximum total severity that a user can accrue from warnings before
+	 * being timeouted in the server.
+	 */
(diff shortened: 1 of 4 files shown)


liquiddevelopmentnet:
I don't think this is a good idea as a mute might not fit well for some warn reasons, I would instead make it so that when you timeout somebody, it will warn the user with severity based on the timeout length if it's not explicitly disabled.

danthe1st:
> I don't think this is a good idea as a mute might not fit well for some warn reasons, I would instead make it so that when you timeout somebody, it will warn the user with severity based on the timeout length if it's not explicitly disabled.

I was generally unsure about whether or not we want this feature for similar reasons.

However, I do think that there should be something that happens _before_ the user is banned from the server.

liquiddevelopmentnet:
> However, I do think that there should be something that happens before the user is banned from the server.

I agree with that, but I still don't think timeouting is a good idea.


> When automatically warning users when they are timouted with a severity based on their timeout duration, I feel like there might be difficulties in choosing proper caps.

I would handle this the same as normal warnings, 20 is minimum 80 is maximum, three severities, LOW(20), MEDIUM(40) and HIGH(80).


> Also, what warn reasons should (not) (necessarily) be count towards timeouts and why?

Example of good reasons : Advertising, Spamming, Racism in chat, Insulting
Example of bad reasons  : Non-english content, Impersonating staff, Repetitively posting JavaScript post in #java-help, DM Advertising

I tried to find real examples from our server that are a good example fit, maybe not all of them are perfect examples.


All in all, I agree that something should happen before the user gets banned, though I think a timeout may not be the best idea, but since I don't have a better idea we should still probably go for it.

danthe1st:
> All in all, I agree that something should happen before the user gets banned, though I think a timeout may not be the best idea, but since I don't have a better idea we should still probably go for it.

I think timeouts are the best way to make the user definitely see they have done something which isn't accepted and that they should stop.

> Example of bad reasons : Non-english content, Impersonating staff, Repetitively posting JavaScript post in #java-help, DM Advertising

If these things happen once, they will probably not get timeouted when doing it the first time (it requires a total severity greater than 50 (by default) for the timeout to get activated).

And the timeout is not a big punishment to be honest (at least by default where it is set to 2h), I would say most of its value is to make sure the user knows to not do that and has a bit of time to think about their actions.

liquiddevelopmentnet (approved):
Seems reasonable

PR #402, opened 2023-03-26
send notification in private thread when DM is unavailable

danthe1st:
When a user disables DM notifications, direct notifications (via `UserNotificationService`) are not sent.

This PR adds changes it so that some notifications are sent in a private thread when DMs are closed.
This should be done for important notifications only.


liquiddevelopmentnet:
I'd add a ping to the user here, so they get notified about the message.

danthe1st:
> I'd add a ping to the user here, so they get notified about the message.

Adding a user to a thread should ping the user.

danthe1st:
Also, note that I have not added QOTW or similar notifications. I wouldn't deem them sufficiently 'important' to ping users who disabled DMs.

PR #405, opened 2023-04-08
change /help-ping to post message requesting help in dedicated channel

danthe1st:
Currently, the `/help-ping` command pings the `help-ping` role.
However, that won't do much as pinging the role doesn't seem to add its members to the thread.

**OLD/OUTDATED version:**
> This PR changes that command so that it adds a random member with the `help-ping` role to the post.
> Aside from that, it makes sure that the command isn't used immediately after post creation.
> 
> I am not sure whether this is the best possible solution but it is the best I could come up with.

**EDIT**:
When users use `/help ping`, a message is sent to a specific channel. This message can be acknowledged (and unacknowledged) by users.

image
image
image

### DIFF
--- a/src/main/java/net/javadiscord/javabot/systems/help/commands/HelpPingSubcommand.java
+++ b/src/main/java/net/javadiscord/javabot/systems/help/commands/HelpPingSubcommand.java
@@ -91,12 +91,12 @@ public class HelpPingSubcommand extends SlashCommand.Subcommand implements Butto
 			lastPingTimes.put(event.getMember().getIdLong(), new Pair<>(System.currentTimeMillis(), config.getGuild()));
 			TextChannel notifChannel = config.getHelpConfig().getHelpNotificationChannel();
 			notifChannel.sendMessageEmbeds(new EmbedBuilder().setDescription("""
-					%s requested help in post %s
+					%s requested help in %s
 					
 					Tags:
 					%s
 					
-					[Post link](%s)
+					[Click to view](%s)
 					"""
 					.formatted(
 							event.getUser().getAsMention(),
--- a/src/main/java/net/javadiscord/javabot/systems/help/commands/HelpPingSubcommand.java
+++ b/src/main/java/net/javadiscord/javabot/systems/help/commands/HelpPingSubcommand.java
@@ -108,7 +108,7 @@ public class HelpPingSubcommand extends SlashCommand.Subcommand implements Butto
 					.setFooter(event.getUser().getId())
 					.setColor(Color.YELLOW)
 					.build())
-				.addActionRow(createResolveButton())
+				.addActionRow(createAcknowledgementButton())
 				.queue();
 			event.reply("""
 					Successfully requested help.
@@ -136,12 +136,12 @@ public class HelpPingSubcommand extends SlashCommand.Subcommand implements Butto
 		return text;
 	}
 
-	private Button createResolveButton() {
-		return Button.of(ButtonStyle.SECONDARY, ComponentIdBuilder.build("help-ping", "acknowledge"), "Mark as resolved");
+	private Button createAcknowledgementButton() {
+		return Button.of(ButtonStyle.SECONDARY, ComponentIdBuilder.build("help-ping", "acknowledge"), "Mark as acknowledged");
 	}
 	
-	private Button createUnresolveButton() {
-		return Button.of(ButtonStyle.SECONDARY, ComponentIdBuilder.build("help-ping", "unacknowledge"), "Mark as unresolved");
+	private Button createUndoAcknowledgementButton() {
+		return Button.of(ButtonStyle.SECONDARY, ComponentIdBuilder.build("help-ping", "unacknowledge"), "Mark as unacknowledged");
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


liquiddevelopmentnet:
I am against this. I know that this is the best solution we know, but i still wouldn't go for it.

Since many people that do **not** actively help also have the help ping role, the command would fail its purpose maybe 75% of the time.

Also, OPs would often then directly address those users added, requesting help from specifically them (since they are kinda "assigned" to that post now), and creating a unnecessary uncomfortable situation for the pinged members.

One approach that came in my mind just now (which maybe isn't a good solution either) would be a channel that pings the help ping role linking the thread where it's requested?

danthe1st:
Yeah, that's also one possibility. Although I'm unsure whether a ping would even be necessary in that case.

Of course, it would also be possible to completely remove that command.

danthe1st:
Yet another possibility would be to add two helpers instead of one and adapt the text shown to the OP to display something like
> Successfully added a random user with the `help-ping` role to the channel. Please note that everyone helping here is doing so voluntarily and they might not be specifically knowledgable in the subject you are having problems with and please respect it if they cannot or do not want to help.

danthe1st:
image
image
image

danthe1st:
Note that this requires to set the config `helpConfig.helpNotificationChannelId`.

src/main/java/net/javadiscord/javabot/systems/help/commands/HelpPingSubcommand.java
```diff
@@ -74,23 +83,65 @@ public void execute(@NotNull SlashCommandInteractionEvent event) {
...
-					.queue();
-			event.replyFormat("Successfully pinged " + role.getAsMention()).setEphemeral(true).queue();
+			TextChannel notifChannel = config.getHelpConfig().getHelpNotificationChannel();
+			notifChannel.sendMessageEmbeds(new EmbedBuilder().setDescription("""
+					%s requested help in post %s
+					
+					Tags:
+					%s
+					
+					[Post link](%s)
```
DynxstyGIT, line 99:
```suggestion
					[Click to view](%s)
```

src/main/java/net/javadiscord/javabot/systems/help/commands/HelpPingSubcommand.java
```diff
@@ -74,23 +83,65 @@ public void execute(@NotNull SlashCommandInteractionEvent event) {
...
 			lastPingTimes.put(event.getMember().getIdLong(), new Pair<>(System.currentTimeMillis(), config.getGuild()));
-			Role role = config.getHelpConfig().getHelpPingRole();
-			event.getChannel().sendMessage(role.getAsMention())
-					.setAllowedMentions(EnumSet.of(Message.MentionType.ROLE))
-					.setEmbeds(buildAuthorEmbed(event.getUser()))
-					.queue();
-			event.replyFormat("Successfully pinged " + role.getAsMention()).setEphemeral(true).queue();
+			TextChannel notifChannel = config.getHelpConfig().getHelpNotificationChannel();
+			notifChannel.sendMessageEmbeds(new EmbedBuilder().setDescription("""
+					%s requested help in post %s
```
DynxstyGIT, line 94:
```suggestion
					%s requested help in %s
```

DynxstyGIT (approved):

DynxstyGIT:
lgtm, maybe turn the tags into a separate embed field tho (and omit it if there are no tags)

liquiddevelopmentnet:
How about creating a self role which allows people to opt into notifications when someone requests help? (Similar to the @help-ping role). Also, whats the mark as resolved button for? Shouldn't it automatically get resolved when the post gets closed?

danthe1st:
I don't think we should implement pinging yet. After all, there are a lot of users with the `help-ping` role.
However, I plan to do the following:
- Make sure the channel is visible by default for existing users with the `help-ping` role (this works by role-restricting the channel at the beginning and then lifting that restriction due to how onboarding works)
- Let the channel be read-only
- Change the self-role in onboarding to a channel selection
- Remove the legacy `help-ping` self-role from the roles channel
- See how that works out (we can add pinging later)

Marking it as resolved means "I will take care of this" so that other people don't need to check. I would for now allow all people to do that and maybe restrict it to helper roles _if necessary_.

liquiddevelopmentnet:
> I don't think we should implement pinging yet. After all, there are a lot of users with the `help-ping` role. However, I plan to do the following:
> 
> * Make sure the channel is visible by default for existing users with the `help-ping` role (this works by role-restricting the channel at the beginning and then lifting that restriction due to how onboarding works)
> * Let the channel be read-only
> * Change the self-role in onboarding to a channel selection
> * Remove the legacy `help-ping` self-role from the roles channel
> * See how that works out (we can add pinging later)
> 
> Marking it as resolved means "I will take care of this" so that other people don't need to check. I would for now allow all people to do that and maybe restrict it to helper roles _if necessary_. The embed keeps a history of the last 5 people who changed that.
> 
> However, we might rename `resolved` to something else in this context. Do you have any suggestions?

Alright, implementing pinging later and first seing if it works sounds fine. Though I dont like the help channel claiming, because:
- It can be easily abused
- It looks bulky (especially if a few people resolve and unresolve the channel) and makes the channel look confusing
- It can easily contain wrong information if someone doesnt manage to find a solution and therefore leave threads unanswered because people think it has been taken care of.
- People will hesitate to use it because it sets an indirect responsibility for the specific thread

I think that the system without the claiming system would work, look and feel better.

danthe1st:
I didn't mean help channels to be "claimed". This is exactly what it is supposed to not do.
I want to allow acknowledging the notification in a way that's visible for others.
Regarding abuse, this is why I implemented the history showing the last few changes in that.
Maybe `acknowledge` is a better name but what would we use for undoing that action? `Unacknowledge` sounds weird so maybe `Retract acknowledgement` or similar?

liquiddevelopmentnet:
> I didn't mean help channels to be "claimed". This is exactly what it is supposed to not do. I want to allow acknowledging the notification in a way that's visible for others. Regarding abuse, this is why I implemented the history showing the last few changes in that. Maybe `acknowledge` is a better name but what would we use for undoing that action? `Unacknowledge` sounds weird so maybe `Retract acknowledgement` or similar?

If its just that, i wouldn't go for it. Doesn't seem like its really needed.

danthe1st:
I don't want many such requests where nobody knows whether someone has looked at them or not. I want to make it as easy as possible for helpers to identify posts in need of help.

liquiddevelopmentnet:
I guess we can give it a try and see what people think of it

liquiddevelopmentnet (approved):

danthe1st:
Should I still change it to `acknowledge` or similar?

liquiddevelopmentnet:
> Should I still change it to `acknowledge` or similar?

Sure

liquiddevelopmentnet (approved):

liquiddevelopmentnet (approved):

liquiddevelopmentnet:
# IMPORTANT BEFORE MERGING:

> Note that this requires to set the config `helpConfig.helpNotificationChannelId`.

PR #407, opened 2023-04-15
change /format-code to add a third option to enable automatic indentation

Fischstaebchenn:
At the moment formartier /format-code messages only as code block. This pull request adds a third option that allows to format the message using a indentation.
Suggestion on the java discord: 1095733157946458222

### DIFF
--- a/src/main/java/net/javadiscord/javabot/systems/user_commands/format_code/FormatCodeCommand.java
+++ b/src/main/java/net/javadiscord/javabot/systems/user_commands/format_code/FormatCodeCommand.java
@@ -1,5 +1,6 @@
 package net.javadiscord.javabot.systems.user_commands.format_code;
 
+import net.javadiscord.javabot.util.*;
 import xyz.dynxsty.dih4jda.interactions.commands.application.SlashCommand;
 import net.dv8tion.jda.api.entities.Message;
 import net.dv8tion.jda.api.events.interaction.command.SlashCommandInteractionEvent;
@@ -9,10 +10,6 @@ import net.dv8tion.jda.api.interactions.commands.build.Commands;
 import net.dv8tion.jda.api.interactions.commands.build.OptionData;
 import net.dv8tion.jda.api.interactions.components.ActionRow;
 import net.dv8tion.jda.api.interactions.components.buttons.Button;
-import net.javadiscord.javabot.util.Checks;
-import net.javadiscord.javabot.util.InteractionUtils;
-import net.javadiscord.javabot.util.Responses;
-import net.javadiscord.javabot.util.StringUtils;
 import org.jetbrains.annotations.Contract;
 import org.jetbrains.annotations.NotNull;
 
@@ -49,7 +46,11 @@ public class FormatCodeCommand extends SlashCommand {
 								.addChoice("SQL", "sql")
 								.addChoice("Swift", "swift")
 								.addChoice("TypeScript", "typescript")
-								.addChoice("XML", "xml")
+								.addChoice("XML", "xml"),
+						new OptionData(OptionType.STRING,"auto-indent","The type of indentation that should be applied to the message, does not automatically indent if left blank",false)
+								.addChoice("Four_Spaces","four")
+								.addChoice("Two_Spaces","two")
+								.addChoice("Tabs","tab")
 				)
 		);
 	}
@@ -64,6 +65,7 @@ public class FormatCodeCommand extends SlashCommand {
 	public void execute(@NotNull SlashCommandInteractionEvent event) {
 		OptionMapping idOption = event.getOption("message-id");
 		String format = event.getOption("format", "java", OptionMapping::getAsString);
+		String indentation = event.getOption("auto-indent","null",OptionMapping::getAsString).toUpperCase();
 		event.deferReply().queue();
 		if (idOption == null) {
 			event.getChannel().getHistory()
@@ -74,7 +76,7 @@ public class FormatCodeCommand extends SlashCommand {
 								.filter(m -> !m.getAuthor().isBot()).findFirst()
 								.orElse(null);
 						if (target != null) {
-							event.getHook().sendMessageFormat("```%s\n%s\n```", format, StringUtils.standardSanitizer().compute(target.getContentRaw()))
+							event.getHook().sendMessageFormat("```%s\n%s\n```", format, IndentationHelper.formatIndentation(StringUtils.standardSanitizer().compute(target.getContentRaw()),IndentationHelper.IndentationType.valueOf(indentation)))
 									.setAllowedMentions(List.of())
 									.setComponents(buildActionRow(target))
 									.queue();
@@ -89,7 +91,7 @@ public class FormatCodeCommand extends SlashCommand {
 			}
 			long messageId = idOption.getAsLong();
 			event.getChannel().retrieveMessageById(messageId).queue(
-					target -> event.getHook().sendMessageFormat("```%s\n%s\n```", format, StringUtils.standardSanitizer().compute(target.getContentRaw()))
+					target -> event.getHook().sendMessageFormat("```%s\n%s\n```", format, IndentationHelper.formatIndentation(StringUtils.standardSanitizer().compute(target.getContentRaw()), IndentationHelper.IndentationType.valueOf(indentation)))
 							.setAllowedMentions(List.of())
 							.setComponents(buildActionRow(target))
 							.queue(),
--- a/src/main/java/net/javadiscord/javabot/systems/user_commands/format_code/FormatCodeCommand.java
+++ b/src/main/java/net/javadiscord/javabot/systems/user_commands/format_code/FormatCodeCommand.java
@@ -48,9 +48,9 @@ public class FormatCodeCommand extends SlashCommand {
 								.addChoice("TypeScript", "typescript")
 								.addChoice("XML", "xml"),
 						new OptionData(OptionType.STRING,"auto-indent","The type of indentation applied to the message, does not automatically indent if left blank.",false)
-								.addChoice("Four_Spaces","four")
-								.addChoice("Two_Spaces","two")
-								.addChoice("Tabs","tab")
+								.addChoice("Four Spaces","FOUR_SPACES")
+								.addChoice("Two Spaces","TWO_SPACES")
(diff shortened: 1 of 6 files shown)


src/test/resources/Unformatted Strings.txt
```diff
@@ -0,0 +1,236 @@
+
```
danthe1st, line 1: Please add a test case with input containing indentation. This should be ignored by the `NULL` formatter and the formatting should be replaced (not added) by the other formatters.

src/test/java/net/javadiscord/javabot/util/IndentationHelperTest.java
```diff
@@ -0,0 +1,44 @@
...
+     */
+    @Test
+    public void testFormatIndentation() {
+        String[] unformatted = null;
+        String[] formatted = null;
+        try {
+            unformatted = Files.readString(Path.of(IndentationHelper.class.getResource("/Unformatted Strings.txt").toURI())).split("----");
+            formatted = Files.readString(Path.of(IndentationHelper.class.getResource("/Formatted Strings.txt").toURI())).split("----");
+        } catch (NullPointerException | URISyntaxException e) {
+            fail("Files to run the test not present");
```
danthe1st, line 27: Why not just throw the exception and let JUnit mark the test case as errored?

danthe1st (changes requested):

src/test/java/net/javadiscord/javabot/util/IndentationHelperTest.java
```diff
@@ -0,0 +1,44 @@
...
+        } catch (IOException e) {
+            fail("IO Exception occurred");
+        }
+
+        try {
+            for(int i = 0,k=0; i< unformatted.length; i++) {
```
danthe1st, line 33: Please use consistent formatting if possible

danthe1st (changes requested):

src/main/java/net/javadiscord/javabot/util/IndentationHelper.java
```diff
@@ -0,0 +1,179 @@
+package net.javadiscord.javabot.util;
```
danthe1st, line 1: Please use tabs for formatting your code. See [the checkstyle failure](https://github.com/Java-Discord/JavaBot/actions/runs/4708493112/jobs/8351238910?pr=407).

src/test/java/net/javadiscord/javabot/util/IndentationHelperTest.java
Fischstaebchenn: I like to give a concrete reason why this test failed. I will pass the throwable as well to provide more information

src/test/resources/Unformatted Strings.txt
Fischstaebchenn: I agree the testcases need to be improved. Would random indentation (between 1 and lets say 32) spaces be enough to determine if not only adding but correcting indentation works correctly ?

danthe1st: It would be sufficient if there was some code with different indentations (tabs, and various amounts of spaces, correct and wrong indentation).

Fischstaebchenn: I will not generate the indentation on the fly by the tester but change the file itself.

src/main/java/net/javadiscord/javabot/systems/user_commands/format_code/FormatCodeCommand.java
```diff
@@ -49,7 +46,11 @@ public FormatCodeCommand() {
 								.addChoice("SQL", "sql")
 								.addChoice("Swift", "swift")
 								.addChoice("TypeScript", "typescript")
-								.addChoice("XML", "xml")
+								.addChoice("XML", "xml"),
+						new OptionData(OptionType.STRING,"auto-indent","The type of indentation that should be applied to the message, does not automatically indent if left blank",false)
```
danthe1st, line 50: This fails at startup as field descriptions can only have 100 characters at maximum.

src/main/java/net/javadiscord/javabot/util/IndentationHelper.java
```diff
@@ -0,0 +1,205 @@
...
+	}
+
+	/**
+	 * Determines if the character in the StringBuilder at the specified position is escaped.
+	 *
+	 * @param builder the StringBuilder which holds the current String
+	 * @param index   The index at which the character to be checked is located.
+	 * @return True, if the escape character is referring to the character at the index, false otherwise.
+	 */
+	private static boolean isEscaped(StringBuilder builder, int index) {
```
danthe1st, line 193: Remove `index` as it is always `builder.length() - 1`.
That would also remove the need for the `index >= builder.length()` check (which would never be executed).
If `index` would be initialized with `builder.length() - 2`, the `index--;` decrement would also not be necessary any more.
With these changes, the `while` loop can be changed to a standard for loop like
```java
for (int index=builder.length() - 2; index > 0 && builder.charAt(index) == '\\'; index--){
    numberOfCharacters++;
}
```

src/main/java/net/javadiscord/javabot/util/IndentationHelper.java
```diff
@@ -0,0 +1,205 @@
...
+					if (current == '\'' && !isEscaped(builder, builder.length() - 1)) {
+						yield IndentationState.CODE;
+					}
+					yield IndentationState.CHARACTER;
+				}
+				case SINGLE_LINE_COMMENT ->
+					current == '\n' ? IndentationState.CODE : IndentationState.SINGLE_LINE_COMMENT;
+				case MULTI_LINE_COMMENT -> {
+					if (current == '*' && i + 1 < text.length() && text.charAt(i + 1) == '/') {
+						yield IndentationState.CODE;
```
danthe1st, line 146: Also increment `i` here and append the next token to the string in order to prevent `*/*` being interpreted as both an end and start of a new comment. Please also add that to a test case.

src/main/java/net/javadiscord/javabot/util/IndentationHelper.java
```diff
@@ -0,0 +1,205 @@
...
+		}
+		IndentationState currentState = IndentationState.CODE;
+		for (int i = 0; i < text.length(); i++) {
+			char current = text.charAt(i);
+			if (startOfLine && current == ' ') {
+				continue;
+			}
+			builder.append(current);
+			if (current == '\n') {
+				builder.append(type.getPattern().repeat(Math.max(numberOfBrackets, 0)));
```
danthe1st, line 124: If there are whitespaces at the start of the line already, are they preserved? Wouldn't that introduce redundant whitespaces? (If yes, please check the code from earlier for a possible mitigation).

danthe1st (changes requested):
Please also add a message context menu command `Format and Indent code` (or similar) which does uses tabs. It should be mostly equivalent to `FormatCodeMessageContext` (except calling `IndentationHelper.formatIndentation(`).

DynxstyGIT:
I mean, something like IntelliJ's "Reformat Code" would be pretty co and useful. This would not only fix the indentation, but also adds missing spaces (e.g. for(int i=1;i>1... => for(int i = 1; i > 1...)
But this is more like gimmicky and I imagine that this would add another layer of complexity without really providing a real benefit besides more readable code

danthe1st:
> I mean, something like IntelliJ's "Reformat Code" would be pretty cool and useful. This would not only fix the indentation, but also adds missing spaces (e.g. for(int i=1;i>1... => for(int i = 1; i > 1...) But this is more like gimmicky and I imagine that this would add another layer of complexity without really providing a real benefit besides more readable code

While that would be possible, I wouldn't want to maintain that (especially with new Java features). Aside from that, there are many different styles.
If we were to (at some point) use something like that, I would rather use a library.

src/main/java/net/javadiscord/javabot/util/IndentationHelper.java
Fischstaebchenn: I would leave it as is for now. Yes it could be simplified but I want to leave it as a utility method as it doesn't readability.

src/main/java/net/javadiscord/javabot/util/IndentationHelper.java
Fischstaebchenn: Whitespaces at the start of a line should not be preserved. If they are please let me know because that is definitely a bug.

danthe1st (approved):

src/main/java/net/javadiscord/javabot/systems/user_commands/format_code/FormatCodeCommand.java
```diff
@@ -49,7 +46,11 @@ public FormatCodeCommand() {
 								.addChoice("SQL", "sql")
 								.addChoice("Swift", "swift")
 								.addChoice("TypeScript", "typescript")
-								.addChoice("XML", "xml")
+								.addChoice("XML", "xml"),
+						new OptionData(OptionType.STRING,"auto-indent","The type of indentation applied to the message, does not automatically indent if left blank.",false)
+								.addChoice("Four_Spaces","four")
```
danthe1st, line 51: These values do not work with `IndentationHelper.IndentationType.valueOf`. They would need to exactly match the enum constants.

danthe1st (changes requested):

src/main/java/net/javadiscord/javabot/util/IndentationHelper.java
danthe1st: Why do you think it doesn't improve readability to replace the whole method with something like
```java
private static boolean isEscaped(StringBuilder builder) {
	for (int index=builder.length() - 2; index > 0 && builder.charAt(index) == '\\'; index--){
    numberOfCharacters++;
	}
	return numberOfCharacters % 2 == 1;
}
```

danthe1st (approved):
Also seems to work when running with JDA

danthe1st:
Is there any specific reason you want to leave the index?

Fischstaebchenn:
For now the index does not really have a purpose but if someone wants to work with this method in the future they might need the index parameter.

DynxstyGIT (approved):
Looks good to me, could use a bit “Reformat Code” tho ;)

liquiddevelopmentnet (approved):

PR #410, opened 2023-05-07
Improve HugListener to accept "fucker", "fuckkkk...", and to use regex instead

0x3C50:


0x3C50:
don't merge yet - "fuckkkk..." doesn't work with uppercase yet

0x3C50:
there we go

src/main/java/net/javadiscord/javabot/listener/HugListener.java
```diff
@@ -24,6 +27,33 @@ public class HugListener extends ListenerAdapter {
 	private final AutoMod autoMod;
 	private final BotConfig botConfig;
 
+	private static final Pattern FUCKER = Pattern.compile("(fuck)(ing|er|k+)?", Pattern.CASE_INSENSITIVE);
+
+	private static String processFuck(String str) {
```
danthe1st, line 32: Please move these methods down so they are together with `processHug` since they belong together.

Also, please make sure that the method names of the `processFuck` and `processHug` methods distinguish what is different between them.

src/main/java/net/javadiscord/javabot/listener/HugListener.java
```diff
@@ -24,6 +27,33 @@ public class HugListener extends ListenerAdapter {
...
+		return FUCKER.matcher(str).replaceAll(matchResult -> {
+			String theFuck = matchResult.group(1);
+			String suffix = Objects.requireNonNullElse(matchResult.group(2), "");
+			String processedSuffix = switch(suffix.toLowerCase()) {
+				case "er", "ing" -> copyCase(suffix, 0, 'g') + suffix;
+				default -> suffix.toLowerCase().startsWith("k") ? copyCase(suffix, "g".repeat(suffix.length())) : "";
```
danthe1st, line 38: I think it would make more sense to use use a distinct `case` for the empty `String`.
```suggestion
				case "" -> "";
				default -> copyCase(suffix, "g".repeat(suffix.length()));
```

src/main/java/net/javadiscord/javabot/listener/HugListener.java
```diff
@@ -24,6 +27,33 @@ public class HugListener extends ListenerAdapter {
...
+			};
+			return processHug(theFuck) + processedSuffix;
+		});
+	}
+
+	private static String copyCase(String source, String toChange) {
```
danthe1st, line 44: Wouldn't it make more sense to create a method that repeats a sequence of a character as often as present in some `String` with the correct case?
e.g. 
```java
private static String repeatWithSameCase(char toRepeat, String cases) {
	StringBuilder ret = new StringBuilder(cases.length());
	for (int i=0; i < cases.length; i++) {
		ret.append(copyCase(cases, i, toRepeat));
	}
	return ret.toString();
}
```

danthe1st (changes requested):

src/main/java/net/javadiscord/javabot/listener/HugListener.java
0x3C50: that'd be oddly specific to make, i think just copying the case makes it a little clearer what actually happens

danthe1st: I'm also OK with your approach but if you want to do that, please use better variable names than `a` and `b`.

src/main/java/net/javadiscord/javabot/listener/HugListener.java
```diff
@@ -55,40 +59,49 @@ public void onMessageReceived(@Nonnull MessageReceivedEvent event) {
...
+				default -> copyCase(suffix, "g".repeat(suffix.length())); // fuckkkkk...
+			};
+			return processHug(theFuck) + processedSuffix;
+		});
+	}
+
+	private static String copyCase(String source, String toChange) {
+		if (source.length() != toChange.length()) throw new IllegalArgumentException("lengths differ");
+		StringBuilder sb = new StringBuilder();
+		for (int i = 0; i < source.length(); i++) {
```
danthe1st, line 94: I think it would make more sense to just replace the loop body with `sb.append(copyCase(source, i, toChange.charAt(i)));`

danthe1st (changes requested):

danthe1st (approved):
Looks good to me, seems to also work when used as the bot.

liquiddevelopmentnet (approved):

src/main/java/net/javadiscord/javabot/listener/HugListener.java
```diff
@@ -80,7 +80,7 @@ private static String replaceFucks(String str) {
 			String theFuck = matchResult.group(1);
 			String suffix = Objects.requireNonNullElse(matchResult.group(2), "");
 			String processedSuffix = switch(suffix.toLowerCase()) {
-				case "er", "ing" -> copyCase(suffix, 0, 'g') + suffix; // fucking, fucker
+				case "er", "ed", "ing" -> copyCase(suffix, 0, 'g') + suffix; // fucking, fucker
```
danthe1st, line 83: The `ed` suffix is not present in the comment.

src/main/java/net/javadiscord/javabot/listener/HugListener.java
```diff
@@ -28,6 +28,17 @@ public class HugListener extends ListenerAdapter {
 	private final AutoMod autoMod;
 	private final BotConfig botConfig;
 
+	public static void main(String[] args) {
```
danthe1st, line 31: Please remove the `main` method again.

danthe1st (changes requested):

danthe1st (approved):

DynxstyGIT (approved):

liquiddevelopmentnet (approved):

PR #414, opened 2023-07-08
XP and DM notification for dormant channels

danthe1st:
### Motivation
From time to time, people think their help channel has been deleted while it is just dormant.

Furthermore, many people looking for help are not closing their post. Helpers should be given some award for all these posts where they are helping without the user closing the post

### Description

This PR adds the following functionality when a help post is marked as dormant:
- The owner of the post gets notified via DM (this can be disabled via a preference)
- Helpers who sent messages get awarded XP
  - Helpers are not awarded with XP when the channel goes dormant multiple times
  - When the post is closed, all helpers with messages get XP, even helpers that have already been awarded due to the channel being dormant.

### Alternatives

It would be possible to not award XP again when the post has been closed. However, this would be a disadvantage to helpers that ask for details, the post goes dormant and help the OP after clarifications/the post is reopened.

It would be possible to remove all messages in the post from the cache once a post goes dormant. This could still be added to the PR.

The owner notification could be disabled by default. However, most people never change their preferences/are not aware and they would still have no idea where their post went.

### Notes

**The migration `02-13-2023_message_cache_links.sql` must be executed once this is deployed!**

It should be verified that the `/db-admin` command works even after the bot restarts before deploying this change.

### DIFF
--- a/src/main/java/net/javadiscord/javabot/systems/help/HelpForumUpdater.java
+++ b/src/main/java/net/javadiscord/javabot/systems/help/HelpForumUpdater.java
@@ -2,13 +2,18 @@ package net.javadiscord.javabot.systems.help;
 
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
+import net.dv8tion.jda.api.EmbedBuilder;
 import net.dv8tion.jda.api.JDA;
 import net.dv8tion.jda.api.entities.Guild;
 import net.dv8tion.jda.api.entities.Message;
+import net.dv8tion.jda.api.entities.MessageEmbed;
 import net.dv8tion.jda.api.entities.channel.concrete.ForumChannel;
 import net.dv8tion.jda.api.entities.channel.concrete.ThreadChannel;
 import net.javadiscord.javabot.data.config.BotConfig;
 import net.javadiscord.javabot.data.config.guild.HelpConfig;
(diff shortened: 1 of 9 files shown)


DynxstyGIT (approved):

src/main/java/net/javadiscord/javabot/systems/help/HelpForumUpdater.java
```diff
@@ -61,13 +68,42 @@ private void checkForumPost(@NotNull ThreadChannel post, HelpConfig config) {
...
+		return new EmbedBuilder()
+				.setTitle("Post marked as dormant")
+				.setDescription(
+						config
+						.getDormantChannelPrivateMessageTemplate()
+						.formatted(
+								post.getAsMention(),
+								post.getParentChannel().getAsMention(),
+								config.getInactivityTimeoutMinutes(),
+								"https://discord.com/channels/" + post.getGuild().getId() + "/" + post.getId()))
```
liquiddevelopmentnet, line 102: Please use [GuildChannel#getJumpUrl](https://ci.dv8tion.net/job/JDA5/javadoc/net/dv8tion/jda/api/entities/GuildChannel.html#getJumpUrl()) instead

liquiddevelopmentnet (changes requested):

liquiddevelopmentnet (approved):

PR #416, opened 2023-07-08
plotting of help XP

danthe1st:
https://canary.discord.com/channels/648956210850299986/1127233526115336344

This PR adds a `plot` argument to the `/help account` command.
If this argument is `true`, the bot attaches a plot containing earned XP values over the last year grouped by month.
Users can only query a plot of their own XP history.

image


DynxstyGIT (approved):

liquiddevelopmentnet:
Wouldn't be a line chart of the total XP over x months more appealing?

danthe1st:
> Wouldn't be a line chart of the total XP over x months more appealing?

Line charts are notmally used when plotting a total changing.
However, this the amount of XP _per month_ and not the total XP the user had each month. The gained XP in month A is independent from the gained XP in month B.

liquiddevelopmentnet (approved):

PR #419, opened 2023-07-12
externalize title of plot, use max for plot height

danthe1st:
Currently, plotting of help XP uses the sum of the gained help XP for computing the height resulting in charts where most of the space is unused.
This PR changes the plotting logic to use the max value instead.
Aside from that, it externalizes the title of the plot for potential reuse of the `Plotter` class.


liquiddevelopmentnet (approved):

danthe1st:
It now also attaches the image _inside_ the embed instead of using an external attachment.

MoonTM-GIT (approved):

PR #426, opened 2023-08-08
fix emoji regulation

danthe1st:
Automated removal of upvote emojis currently has the following two problems:
- For fallback emojis, it removes the suggestion emoji sent by the bot itself
- It does not recognize custom emojis because [`CustomEmoji` and `RichCustomEmoji` violate symmetricity of the `equals` contract](https://github.com/discord-jda/JDA/pull/2519).
  - `getUpvoteEmoji(event)` returns a `RichCustomEmoji` which is a subclass of `CustomEmoji` while `event.getEmoji()` returns a `CustomEmoji` that is not a `RichCustomEmoji`

This PR fixes these issues.


liquiddevelopmentnet (approved):

danthe1st:
Since https://github.com/discord-jda/JDA/pull/2519 is now merged, this would also be fixed when updating to the next JDA version once it comes out.

MoonTM-GIT (approved):

PR #429, opened 2023-08-10
add timestamp to acknowledging help notifications

danthe1st:
This PR adds a timestamp to the embed field stating users acknowledging or unacknowledging a help channel notification (created by `/help ping`).

image

Furthermore, it changes the description to not include the `help-ping` role.

### DIFF
--- a/src/main/java/net/javadiscord/javabot/listener/HugListener.java
+++ b/src/main/java/net/javadiscord/javabot/listener/HugListener.java
@@ -1,8 +1,6 @@
 package net.javadiscord.javabot.listener;
 
 import lombok.RequiredArgsConstructor;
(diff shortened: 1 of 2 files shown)


liquiddevelopmentnet:
How about using the relative (R) timestamp format? It would make it look more like a log and would improve readability.

MoonTM-GIT (approved):

Pleezon:
> unsure whether relative timestamps are really better here.
Personally, I think that a relative timestamp would fit better since people checking the channel would primarily want to know how recent the last acknowledgments were.

danthe1st:
image
It's using relative timestamps now.

Pleezon (approved):

PR #430, opened 2023-08-15
staff activity tracking

danthe1st:
This PR adds tracking of staff activity.

A message is maintained for each staff member. This message contains an embed with the timestamp of the last sent message of the staff member.
image

### Merge Notes / Dependencies
This PR comes with the migration `08-15-2023_staff_activity.sql`.
This migration should be executed when deploying this change.
~~Since the `/db-admin` currently has issues, this should be deployed with or after #422.~~ This has been merged

### DIFF
--- a/src/main/java/net/javadiscord/javabot/data/config/guild/ModerationConfig.java
+++ b/src/main/java/net/javadiscord/javabot/data/config/guild/ModerationConfig.java
@@ -34,6 +34,11 @@ public class ModerationConfig extends GuildConfigItem {
 	 * The threshold for deleting a message in #share-knowledge. Note that this should be strictly < 0.
 	 */
 	private int shareKnowledgeMessageDeleteThreshold;
+	
+	/**
+	 * ID of the channel storing staff activity information.
+	 */
+	private long staffActivityChannelId = 0;
 
 	/**
 	 * The threshold for deleting a message in #looking-for-programmer.
@@ -104,6 +109,10 @@ public class ModerationConfig extends GuildConfigItem {
 
 	public ForumChannel getJobChannel() {
 		return this.getGuild().getForumChannelById(this.jobChannelId);
+	}
+	
+	public TextChannel getStaffActivityChannel() {
+		return this.getGuild().getTextChannelById(this.staffActivityChannelId);
 	}
 
 	public ForumChannel getShareKnowledgeChannel() {
--- a/src/main/java/net/javadiscord/javabot/systems/staff_activity/StaffActivityListener.java
+++ b/src/main/java/net/javadiscord/javabot/systems/staff_activity/StaffActivityListener.java
@@ -1,0 +1,74 @@
+package net.javadiscord.javabot.systems.staff_activity;
+
+import lombok.RequiredArgsConstructor;
+import net.dv8tion.jda.api.EmbedBuilder;
+import net.dv8tion.jda.api.entities.Member;
+import net.dv8tion.jda.api.entities.MessageEmbed;
+import net.dv8tion.jda.api.entities.channel.concrete.TextChannel;
+import net.dv8tion.jda.api.events.message.MessageReceivedEvent;
+import net.dv8tion.jda.api.hooks.ListenerAdapter;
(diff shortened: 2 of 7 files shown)


liquiddevelopmentnet:
Branch name lol.

I think it would be better, displaying all staff members in one embed, since its only one line anyways. Otherwise it would be hard to get a quick overview.
Also, the lines could be sorted by last time a message was sent.

Example:

@Userondiscord1 | Last message: 2 years ago
@Userondiscord2 | Last message: 3 hours ago
@Userondiscord3 | Last message: 5 seconds ago

danthe1st:
There is a full channel available for that feature so I don't really see a reason to require a single message.
Also, having a message per user allows to easily delete these messages in case the moderator resigns.
Aside from that, it might be hard to read if we want to have more information per user (e.g. user ID, mention, last message timestamp, last command timestamp etc.). This only gets worse on mobiles/small devices.
In case of splitting it over multiple lines with formatting, it could be difficult with the message length limit.

liquiddevelopmentnet:
> There is a full channel available for that feature so I don't really see a reason to require a single message.
A full channel is no reason to make it messy

> Also, having a message per user allows to easily delete these messages in case the moderator resigns.
The Bot would remove a user since they don't have the Staff role anymore

> Aside from that, it might be hard to read if we want to have more information per user...
Yes, that is true. But currently your implementation just contains the last message sent, which would be better organized in one message all together.

If we will add more fields to this later (possibly also buttons?) we can leave it like this though, i just thought the way it now is doesn't require a full message.

danthe1st:
> A full channel is no reason to make it messy
I think putting everything in a single message is messoer than one message per user.  
> The Bot would remove a user since they don't have the Staff role anymore  
That would require the bot ro be able to reliably detect it (and reliably detect any inconsistencies and automatically remove them without removing any "correct" parts).
> Yes, that is true. But currently your implementation just contains the last message sent, which would be better organized in one message all together.  
Would that be organized together? After all, it does contact the user ID, effective name (both necessary in case Discord doesn't load it), mention and timestamp which is a bit much for a single line.
Also, single lines do not allow including the profile icon but this is a minor thing that doesn't really matter.

danthe1st:
> A full channel is no reason to make it messy
I think putting everything in a single message is messoer than one message per user.  
> The Bot would remove a user since they don't have the Staff role anymore  
That would require the bot ro be able to reliably detect it (and reliably detect any inconsistencies and automatically remove them without removing any "correct" parts).
> Yes, that is true. But currently your implementation just contains the last message sent, which would be better organized in one message all together.  
Would that be organized together? After all, it does contact the user ID, effective name (both necessary in case Discord doesn't load it), mention and timestamp which is a bit much for a single line.
Also, single lines do not allow including the profile icon but this is a minor thing that doesn't really matter.

However, it might be a good idea to use embed fields and already include the timestamp of the last 'moderative action' (i.e. bot interaction like warn/timeout/kick/ban).

liquiddevelopmentnet:
I just don't like the idea of having a channel full of unorganized messages.

liquiddevelopmentnet (approved):

Pleezon:
An alternative approach could be displaying a staff member's status as *active* until they haven't sent a message for more than e.g. 7 days. I'd find it more applicable because I believe it accomplishes the system's goal of detecting general inactivity more than something that keeps track of detailed activity in order to conclude possible inactivity.
Just throwing this in as a thought - I am fully content with the current implementation.

DynxstyGIT:
lmao the branch name
lgtm

DynxstyGIT (approved):

danthe1st:
> An alternative approach could be displaying a staff member's status as _active_ until they haven't sent a message for more than e.g. 7 days. I'd find it more applicable because I believe it accomplishes the system's goal of detecting general inactivity more than something that keeps track of detailed activity in order to conclude possible inactivity. Just throwing this in as a thought - I am fully content with the current implementation.

I don't think we need a boolean active/inactive indicator. Also, maintaining that would require changing that regularly or using a timer and it would be more complicated if we wanted to add more metrics.

PR #425, opened 2023-08-07
AutoCodeFormatter

Pleezon:
- added AutoCodeFormatter.java - a class focused on implementing functionality to remind users of using code syntax highlighting in help channels.
Currently, the system uses "{" and "}" to detect a codeblock that isnt formatted correctly.
- added a preference for enabling/disabling AutoCodeFormatter
- added a Setting in HelpConfig: hint for when users don't format their code
- added a Setting in HelpConfig: footnote for when code has been autoformatted
- edited HelpListener to include a call to AutoCodeFormatter.java
- moved replaceMemberMessage from HugListener to WebhookUtil to allow usage from within AutoCodeFormatter.java

### Merge/deploy note
This should only be deployed with/after https://github.com/Java-Discord/JavaBot/pull/422 because it uses preferences which are currently broken (the command is not registered).

PR #436, opened 2023-09-10
improve detection of codeblocks for autoformat

danthe1st:
The bot attempts to automatically format messages containing code using codeblocks (#425).

This PR changes that behaviour so that messages are not formatted when the first `{` and last `}` are on the same line and also not if a message contains even a single backtick (instead of three).

The former is because things like `{something}` probably don't need codeblocks and the latter is due to the assumption that messages containing backticks are probably formatted in some way. If they already formatted a part of the code, they might want their formatting to be used.

See also https://canary.discord.com/channels/648956210850299986/1150336575675318272 and probably also similar threads where the bot formatted messages that shouldn't be changed.


Pleezon (approved):
lgtm, was needed

MoonTM-GIT (approved):

PR #437, opened 2023-09-10
add optional comments to /help ping

danthe1st:
This PR allows to add a comment to `/help ping` explaining why specifically they used the command.

This could allow users to differentiate cases where nobody responds from cases where the helpers in the post are not able to solve the issue.
I am not sure whether we should actually do this as I don't know how people would use this.

image


danthe1st:
This could allow users to differentiate cases where nobody responds from cases where the helpers in the post are not able to solve the issue.
I am not sure whether we should actually do this as I don't know how people would use this.

DynxstyGIT (approved):

MoonTM-GIT:
> I am not sure whether we should actually do this as I don't know how people would use this.

Since it's optional I don't see why not

PR #440, opened 2023-09-12
numeric metrics

danthe1st:
The `/guild/{guild_id}/metrics` endpoint currently uses `String`s for metrics.

This PR changes both the `metricConfig` and the DTO for that endpoint to use `long` instead of `String`.

# Important (deploy) note
⚠️ THIS PR CONTAINS A (POTENTIALLY) BACKWARDS INCOMPATIBLE GUILD CONFIG CHANGE! ⚠️

If no elements are stored in the `weeklyMessages`/`activeMembers` of the guild config, the guild config cannot be read.
Therefore, the following should be done **before deploying**
- Create a backup of the guild config.
- Ensure that `metricsConfig.weeklyMessages` and `metricsConfig.activeMembers` are **not** empty `String`s.

### DIFF
--- a/src/main/java/net/javadiscord/javabot/data/config/guild/MetricsConfig.java
+++ b/src/main/java/net/javadiscord/javabot/data/config/guild/MetricsConfig.java
@@ -11,8 +11,8 @@ import net.javadiscord.javabot.data.config.GuildConfigItem;
 @Data
 @EqualsAndHashCode(callSuper = true)
 public class MetricsConfig extends GuildConfigItem {
-	private String weeklyMessages = "";
-	private String activeMembers = "";
+	private long weeklyMessages = -1;
+	private long activeMembers = -1;
 	private long metricsCategoryId = 0;
 	private String metricsMessageTemplate = "";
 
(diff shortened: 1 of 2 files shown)


DynxstyGIT (approved):

Pleezon (approved):
lgtm

src/main/java/net/javadiscord/javabot/data/config/guild/MetricsConfig.java
```diff
@@ -11,8 +11,8 @@
 @Data
 @EqualsAndHashCode(callSuper = true)
 public class MetricsConfig extends GuildConfigItem {
-	private String weeklyMessages = "";
-	private String activeMembers = "";
+	private long weeklyMessages = -1;
+	private long activeMembers = -1;
```
Pleezon, line 15: Side question: why are these initialized with -1 rather than 0?

danthe1st: If they are -1,it can clearly be differentiated as "no value present" asopposed to "no messages".

DynxstyGIT:
Is there anything keeping this from being merged?

danthe1st:
@DynxstyGIT 
> Is there anything keeping this from being merged? 

Not really but care should be taken when deploying as mentioned in the merge note.

PR #442, opened 2023-09-17
Config improvements

danthe1st:
This PR improves the `/config` command:
- It adds autocomplete to `/config set` and `/config get`
- It makes the `value` option of `/config set` optional and allows using a modal to set this value
- It removes the restriction that only bot admin users can use `/config` as this command only allows updating the guild config. With this, guild admins can edit the guild config.


DynxstyGIT (approved):
good stuff

MoonTM-GIT (approved):

PR #449, opened 2023-10-14
Include embeds in starboard and merge images with starboard embeds

Defective4:
This commit makes JavaBot include embeds in starboard messages.
The bot will also merge image attachment from the starred mesasge into the starboard message embed.
This works **only** if the message has only one attachment, otherwise it falls back to the old behavior.
If starred message contains embeds, all fields of message's *first* embed are included in the starboard message.
It might break when a message contains both an embed with image and exactly one attachment though.

### DIFF
--- a/src/main/java/net/javadiscord/javabot/systems/starboard/StarboardManager.java
+++ b/src/main/java/net/javadiscord/javabot/systems/starboard/StarboardManager.java
@@ -149,6 +149,10 @@ public class StarboardManager extends ListenerAdapter {
 					action.addFiles(FileUpload.fromData(a.getProxy().download().get(), a.getFileName()));
 				} catch (InterruptedException | ExecutionException e) {
 					action.addContent("Could not add Attachment: " + a.getFileName());
+					ExceptionLogger.capture(e);
+					if(e instanceof InterruptedException) {
+						Thread.currentThread().interrupt();
+					}
 				}
 			}
 		}
@@ -231,8 +235,6 @@ public class StarboardManager extends ListenerAdapter {
 				.setDescription(message.getContentRaw())
 				.setFooter("#" + message.getChannel().getName());
 		
-		boolean attachedImage = false;
-		
 		List<MessageEmbed> embeds = message.getEmbeds();
 		if(embeds.size() > 0) {
 			MessageEmbed firstEmbed = embeds.get(0);
@@ -242,15 +244,12 @@ public class StarboardManager extends ListenerAdapter {
 			ImageInfo image = firstEmbed.getImage();
 			if(image != null && image.getUrl() != null) {
 				builder.setImage(image.getUrl());
-				attachedImage = true;
 			}
 		}
 		
-		if(!attachedImage) {
-			List<Attachment> attachments = message.getAttachments();
-			if(attachments.size() == 1) {
-				builder.setImage(attachments.get(0).getUrl());
-			}
+		List<Attachment> attachments = message.getAttachments();
+		if(attachments.size() == 1) {
+			builder.setImage(attachments.get(0).getUrl());
 		}
 		
 		return builder.build();
--- a/src/main/java/net/javadiscord/javabot/systems/starboard/StarboardManager.java
+++ b/src/main/java/net/javadiscord/javabot/systems/starboard/StarboardManager.java
@@ -4,6 +4,9 @@ import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import net.dv8tion.jda.api.EmbedBuilder;
 import net.dv8tion.jda.api.entities.*;
+import net.dv8tion.jda.api.entities.Message.Attachment;
+import net.dv8tion.jda.api.entities.MessageEmbed.Field;
+import net.dv8tion.jda.api.entities.MessageEmbed.ImageInfo;
 import net.dv8tion.jda.api.entities.channel.ChannelType;
 import net.dv8tion.jda.api.entities.channel.middleman.MessageChannel;
 import net.dv8tion.jda.api.entities.emoji.Emoji;
@@ -25,6 +28,7 @@ import net.javadiscord.javabot.util.UserUtils;
 import org.jetbrains.annotations.NotNull;
 import org.springframework.dao.DataAccessException;
 
+import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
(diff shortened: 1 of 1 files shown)


src/main/java/net/javadiscord/javabot/systems/starboard/StarboardManager.java
```diff
@@ -137,13 +141,18 @@ private void addMessageToStarboard(Message message, int stars, @NotNull Starboar
...
-			} catch (InterruptedException | ExecutionException e) {
-				action.addContent("Could not add Attachment: " + a.getFileName());
+		List<Attachment> attachments = message.getAttachments();
+		
+		if(attachments.size() > 1) {
+			for (Message.Attachment a : attachments) {
+				try {
+					action.addFiles(FileUpload.fromData(a.getProxy().download().get(), a.getFileName()));
+				} catch (InterruptedException | ExecutionException e) {
+					action.addContent("Could not add Attachment: " + a.getFileName());
```
danthe1st, line 151: Re-interrupt the thread when an `InterruptedException` occurs.

src/main/java/net/javadiscord/javabot/systems/starboard/StarboardManager.java
```diff
@@ -137,13 +141,18 @@ private void addMessageToStarboard(Message message, int stars, @NotNull Starboar
...
-			try {
-				action.addFiles(FileUpload.fromData(a.getProxy().download().get(), a.getFileName()));
-			} catch (InterruptedException | ExecutionException e) {
-				action.addContent("Could not add Attachment: " + a.getFileName());
+		List<Attachment> attachments = message.getAttachments();
+		
+		if(attachments.size() > 1) {
+			for (Message.Attachment a : attachments) {
+				try {
+					action.addFiles(FileUpload.fromData(a.getProxy().download().get(), a.getFileName()));
```
danthe1st, line 149: If possible, please don't download and re-upload the image but instead use the existing URL.

Also, don't block the thread for network requests.

src/main/java/net/javadiscord/javabot/systems/starboard/StarboardManager.java
```diff
@@ -216,11 +225,34 @@ private boolean removeMessageFromStarboard(long messageId, MessageChannel channe
...
+			if(image != null && image.getUrl() != null) {
+				builder.setImage(image.getUrl());
+				attachedImage = true;
+			}
+		}
+		
+		if(!attachedImage) {
+			List<Attachment> attachments = message.getAttachments();
+			if(attachments.size() == 1) {
+				builder.setImage(attachments.get(0).getUrl());
```
danthe1st, line 252: Doesn't that result in the image being present twice (once as a normal attachment and once as an embed attachment) if you don't use an `attachment://` URL?

danthe1st:
https://canary.discord.com/channels/648956210850299986/752535909228085348/1162722901045289010

danthe1st (changes requested):
It is not fully decided whether this feature is wanted.
However, I do agree with adding this.

src/main/java/net/javadiscord/javabot/systems/starboard/StarboardManager.java
Defective4: It won't. That's why I added a check [here](https://github.com/Defective4/JavaBot/blob/60529d695b853cf1663b6f784eee3ff2421bb921/src/main/java/net/javadiscord/javabot/systems/starboard/StarboardManager.java#L146)
The image won't  attached if the original message has only one attachment

src/main/java/net/javadiscord/javabot/systems/starboard/StarboardManager.java
Defective4: Actually the file upload is **not** my code. I just added the `if` here

src/main/java/net/javadiscord/javabot/systems/starboard/StarboardManager.java
Defective4: https://github.com/Java-Discord/JavaBot/pull/449#discussion_r1359447444

src/main/java/net/javadiscord/javabot/systems/starboard/StarboardManager.java
danthe1st: I see. Yeah, that would break if there is a single normal attachment and an embed attachment.

src/main/java/net/javadiscord/javabot/systems/starboard/StarboardManager.java
Defective4: > If possible, please don't download and re-upload the image but instead use the existing URL.
> 
> Also, don't block the thread for network requests.

Also I just checked. It's not possible to NOT reupload an attachment if you want to include it along a embed. If you just send an URL Discord won't convert it to an image.
image

src/main/java/net/javadiscord/javabot/systems/starboard/StarboardManager.java
danthe1st: Can you please still extract the `InterruptedException` catch block and re-interrupt the thread using `Thread.currentThread().interrupt()` as well use `ExceptionLogger.capture` for saving the exeption?

src/main/java/net/javadiscord/javabot/systems/starboard/StarboardManager.java
Defective4: > and an embed attachment.

Yeah, only if the embed also has an image but yes, I'm aware of that and I don't know how to work around that

danthe1st (changes requested):

src/main/java/net/javadiscord/javabot/systems/starboard/StarboardManager.java
danthe1st: You could just use an `attachment://` URL (`"attachment://"+fileUpload.getName()`) or don't skip adding the attachment in case `attachedImage` is true (parts of that code could be moved up before the reupload step)

danthe1st (changes requested):
I think that embed descriptions should be copied as well.
However, if a message contains both text content and an embed with a description, only the text content should be used as an embed description.

danthe1st (changes requested):

src/main/java/net/javadiscord/javabot/systems/starboard/StarboardManager.java
```diff
@@ -245,6 +245,9 @@ private boolean removeMessageFromStarboard(long messageId, MessageChannel channe
 			if(image != null && image.getUrl() != null) {
 				builder.setImage(image.getUrl());
 			}
+			String desc = message.getContentRaw();
+			if(desc == null || desc.isEmpty())
```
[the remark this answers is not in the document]
Defective4: Already fixed in commit [a6ef055](https://github.com/Java-Discord/JavaBot/pull/449/commits/a6ef0553167c854d127174d4f193a457b71e1153).
Sorry about the mistake!

danthe1st (commented):
Implementation-wise, it looks good and seems to work.
However, I would like to wait for further feedback on what other people think about this before approving.

DynxstyGIT (approved):

danthe1st (approved):

PR #450, opened 2023-10-16
awarded help XP distribution statistics

danthe1st:
This PR adds a `/help stats` command showing the XP distribution of different users over the last 12 months.

image

It also changes `HelpAccount` to not encapsulate the `BotConfig` instance in order to reuse it for experience.

https://canary.discord.com/channels/648956210850299986/752535909228085348/1162749758327230555

It also disables the check that `/help account plot:True` can only be used on the own user.

### Considered alternatives
- If the bot used Java 21, `LinkedHashMap#reversed` could have been used instead of adding and populating a `topUsers` list.
- It would have been possible to use a different class/record instead of `HelpAccount` to store XP information per user. However, the `HelpAccount` class seems to fit that purpose.
- legend/axis information in image
  - Since this would require drawing more content on the image which would be fragile and might overlap with parts of the plot, the embed was used for that information.
- using usernames/displaynames instead of mentions in the embed
  - This would require fetching the top users from Discord's API and implementing a fallback strategy in case the user is not available any more (deleted account, left server, etc). It is still possible to change that.
- It would have been possible to always give a colour to the user executing the command by adding that user to the `LinkedHashMap` at the beginning.
- use help leaderboard data for determining users with colours
  - This would include XP decay and users who where awarded a lot of help XP in the given time period might be different from users who currently have the most help XP.
- Keep the restriction that `/help account plot:True` only works with the current user.
- Add some more statistics to this embed
  - Which statistics would be interesting?

### DIFF
--- a/src/main/java/net/javadiscord/javabot/api/routes/user_profile/UserProfileController.java
+++ b/src/main/java/net/javadiscord/javabot/api/routes/user_profile/UserProfileController.java
@@ -109,7 +109,7 @@ public class UserProfileController extends CaffeineCache<Pair<Long, Long>, UserP
 				data.setQotwAccount(qotwAccount);
 				// Help Account
 				HelpAccount helpAccount = helpExperienceService.getOrCreateAccount(user.getIdLong());
-				data.setHelpAccount(HelpAccountData.of(helpAccount, guild));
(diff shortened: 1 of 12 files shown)


DynxstyGIT (approved):
oh that's cool, lgtm!

MoonTM-GIT (approved):

PR #463, opened 2023-12-29
create docker image via CI

danthe1st:
This PR creates a docker image using a Java 21 JRE automatically on commits to `main`.

Specifically, it contains the following changes:
- Gradle and Lombok were updated to work with Java 21 (for a future update)
- The default CI script was changed to run on every branch
- A new CI script was added.  
  This script builds a docker image that allows running the bot.

The docker image can be executed like that (where `USERNAME/REPONAME` is the repository of the docker image, e.g. `dan1st/javabot-test`)
```bash
docker run --rm -v "$PWD/config:/work/config" -v "$PWD/logs:/work/logs" -v "$PWD/db:/work/db" -v "$PWD/purgeArchives:/work/purgeArchives" -p 9000:9000 -p 9001:9001 USERNAME/REPONAME
```

This command forwards both the HTTP and AJP port.

Publishing the image to Docker Hub requires the following secrets to be present:
- `DOCKERHUB_USERNAME`: The username (or organization name) for Docker Hub, e.g. `dan1st`
- `DOCKERHUB_REPONAME`: The repository name for Docker Hub, e.g. `javabot`
- `DOCKERHUB_TOKEN` The Docker Hub token for uploading

If `DOCKERHUB_USERNAME` is missing, the image will not be uploaded but the build should still work.

For testing, a repository is available [here](https://hub.docker.com/r/dan1st/javabot-test).  
An example run can be found [here](https://github.com/danthe1st/JavaBot/actions/runs/7369415856)

# Alternatives
- ~~It would be possible to use one script for both verifying the build and uploading the Docker image and only build the image on the main branch. This way, there wouldn't be 2 similar scripts running on every push to `main`.~~ This was implemented.
- It would be possible to require Java 21 for building. This can be done in a future PR.
- A different base image/JRE vendor could be used instead Eclipse Adoptium/Temurin.
- It would be possible to use GraalVM's `native-image` for a Docker image not requiring a JRE. This was not done due to it being pretty fragile in my experience/quite some work to maintain (I don't want to write more bugreports to Spring/etc. due to issues with that).
- It would be possible to use `jlink` to create the JRE.
- It may be a good idea to change `server.address=127.0.0.1` to `server.address=0.0.0.0` in the `application.properties`. See the merge notes.

# Merge notes - check before merging/deploying/using
- ~~The branch `main` should be protected! Only PRs should be allowed. Make sure this is the case.~~ This seems to be the case already.
- Using this requires changes to the deployment process.  
For example, it would be possible to execute `docker pull USERNAME/REPONAME` and the aforementioned `docker run` command in order to (re)start the bot  
This could be done every time the bot is stopped with `/redeploy`.
- This requires the DB to be located in a directory called `db` in the directory where the `docker run` command is executed.
  - This requires moving the DB and changing the `systems.json` config.
- a full backup of the bot should be created before deploying!
  - It might also be a good idea to _automatically_ create a backup of the config and DB after every restart (or periodically) if this isn't done already.
- To make the API accessible, it is necessary to add an `application.properties` with the content `server.address=0.0.0.0` to the `config` directory. Something similar may be necessary for AJP though it seems to be accessible from outside anyways. If AJP is used without requiring the HTTP endpoint, this is not necessary.


danthe1st:
This is an alternative to the effort in #330

danthe1st:
We might also want to update the `README` to add instructions for this. I can add that to the PR if necessary/wanted.

MoonTM-GIT (approved):

jasonlessenich (approved):

PR #470, opened 2024-01-03
Fixed some first-time startup issues

jasonlessenich:
This PR fixes several issues when firing up the bot for the first time:
- Prevents the bot from initializing the database when no JDA Token is set
- Adds a default value for the AJP Secret (bot would not startup without. Default is `secret`)
- Filters out queries that contain comments when initializing the database schema (Would simply skip those queries, resulting in missing tables)
- Added missing semicolon and fixed primary key in database schema
- (Added `IF NOT EXISTS` to table statements)
- (Added note about bot intents in readme)

### DIFF
--- a/src/main/java/net/discordjug/javabot/SpringConfig.java
+++ b/src/main/java/net/discordjug/javabot/SpringConfig.java
@@ -43,8 +43,9 @@ public class SpringConfig {
 
 	@Bean
 	DataSource dataSource(BotConfig config) {
-		if (config.getSystems().getJdaBotToken().isEmpty())
+		if (config.getSystems().getJdaBotToken().isEmpty()) {
 			throw new RuntimeException("JDA Token not set. Stopping Bot...");
+		}
 		return DbHelper.initDataSource(config);
 	}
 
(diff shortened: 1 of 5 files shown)


src/main/java/net/discordjug/javabot/SpringConfig.java
```diff
@@ -43,7 +42,9 @@ PresenceUpdater standardActivityPresenceUpdater() {
 	}
 
 	@Bean
-	DataSource getDataSource(BotConfig config) {
+	DataSource dataSource(BotConfig config) {
+		if (config.getSystems().getJdaBotToken().isEmpty())
+			throw new RuntimeException("JDA Token not set. Stopping Bot...");
```
danthe1st, line 47: checkstyle issue here due to missing curley brackets

danthe1st (changes requested):

danthe1st (approved):

danthe1st:
It might still be a good idea to create a DB export without data and use the table definitions from that as the schema as the current schema might not be completely up-to-date due to (possible) missing migrations.

PR #473, opened 2024-01-07
use java codeblock when autoformatting code in help channels

danthe1st:
Currently, automatically created codeblocks in help channels (introduced in #425) do not use `java` syntax highlighting.

This PR automatically adds `java` syntax highlighting to all codeblocks that are created automatically when a message in a help channel looks like code.

Suggestion: https://canary.discord.com/channels/648956210850299986/1193480663064662046


jasonlessenich (approved):

Pleezon (approved):
lgtm - Although I believe we should probably create a global variable (or use JDA) for discord's character limit since it might vary.

danthe1st:
> lgtm - Although I believe we should probably create a global variable (or use JDA) for discord's character limit since it might vary.

I used `Message.MAX_CONTENT_LENGTH` for this now.

PR #474, opened 2024-01-09
restrict closing of posts by non-original posters

danthe1st:
This PR adds the following restrictions when users other than the original poster close help posts:
- The action is logged in the log channel
- Closing via buttons is prevented.
- A reason is required.

This does not affect users closing their own posts.

Log message:
grafik
Closing without specifying a reason:
grafik
Closing using buttons (same message if user has permissions to mod-close the post or doesn't have these permissions):
grafik

### DIFF
--- a/src/main/java/net/discordjug/javabot/Bot.java
+++ b/src/main/java/net/discordjug/javabot/Bot.java
@@ -112,10 +112,12 @@ public class Bot implements ApplicationListener<ApplicationReadyEvent> {
 		List<IdMapping<StringSelectMenuHandler>> stringSelectMappings = new ArrayList<>();
 		for (Object handler : interactionHandlers.values()) {
 			AutoDetectableComponentHandler annotation = handler.getClass().getAnnotation(AutoDetectableComponentHandler.class);
-			String[] keys = annotation.value();
-			addComponentHandler(buttonMappings, keys, handler, ButtonHandler.class);
-			addComponentHandler(modalMappings, keys, handler, ModalHandler.class);
-			addComponentHandler(stringSelectMappings, keys, handler, StringSelectMenuHandler.class);
+			if (annotation != null) {//superclasses are annotated, ignore
+				String[] keys = annotation.value();
+				addComponentHandler(buttonMappings, keys, handler, ButtonHandler.class);
+				addComponentHandler(modalMappings, keys, handler, ModalHandler.class);
+				addComponentHandler(stringSelectMappings, keys, handler, StringSelectMenuHandler.class);
+			}
 		}
 		dih4jda.addButtonMappings(buttonMappings.toArray(IdMapping[]::new));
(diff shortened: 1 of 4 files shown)


liquiddevelopmentnet (changes requested):
This looks good in general, however, I think that displaying a modal when somebody tries to close a help post from another user which is non staff is better because since the field isn't required by discord, they probably won't supply a reason at the first time. That means the command would always has to be run twice. Another reason is that the reason field may get confused because it "technically" has different use cases between certified helpers, mods, and OPs.

I'm aware these are just accessibility improvements, but I think this is worth a discussion.

danthe1st:
> This looks good in general, however, I think that displaying a modal when somebody tries to close a help post from another user which is non staff is better because since the field isn't required by discord, they probably won't supply a reason at the first time. That means the command would always has to be run twice. Another reason is that the reason field may get confused because it "technically" has different use cases between certified helpers, mods, and OPs.
> 
> I'm aware these are just accessibility improvements, but I think this is worth a discussion.

To be honest, the original poster supplying a reason is fairly pointless.

While a modal would be possible, I don't see much of a difference since people can just type the command again.

liquiddevelopmentnet:
> > This looks good in general, however, I think that displaying a modal when somebody tries to close a help post from another user which is non staff is better because since the field isn't required by discord, they probably won't supply a reason at the first time. That means the command would always has to be run twice. Another reason is that the reason field may get confused because it "technically" has different use cases between certified helpers, mods, and OPs.
> > I'm aware these are just accessibility improvements, but I think this is worth a discussion.
> 
> To be honest, the original poster supplying a reason is fairly pointless.
> 
> While a modal would be possible, I don't see much of a difference since people can just type the command again.

As i said, mainly accessibility , your decision in the end

liquiddevelopmentnet (approved):

danthe1st:
> > > This looks good in general, however, I think that displaying a modal when somebody tries to close a help post from another user which is non staff is better because since the field isn't required by discord, they probably won't supply a reason at the first time. That means the command would always has to be run twice. Another reason is that the reason field may get confused because it "technically" has different use cases between certified helpers, mods, and OPs.
> > > I'm aware these are just accessibility improvements, but I think this is worth a discussion.
> > 
> > 
> > To be honest, the original poster supplying a reason is fairly pointless.
> > While a modal would be possible, I don't see much of a difference since people can just type the command again.
> 
> As i said, mainly accessibility , your decision in the end

This is now implemented.
Closing (posts of other people) via buttons is still disabled in order to make it "easier" to look whether closing the post would interrupt someone helping.

Also, I have configured the modal to require at least 11 characters in order to prevent overly lazy texts (e.g. `JavaScript` since that's not really informative to the author). This restriction is currently not in place for commands.

liquiddevelopmentnet (approved):
lgtm!

MoonTM-GIT (approved):

PR #479, opened 2024-02-12
warn decay

danthe1st:
This PR adds functionality to slowly decrease the severtiy of warns over time.

The severity weight subtraction removed does not scale with the severity of the warns.
Having more/higher-severity warns does not mean that the severity is reduced more.

It also includes a refactoring of `ModerationService` to be a Spring-managed bean/service class.

Furthermore, it changes `/warn discard-all` to only discard potentially active warns (i.e. warns issues within the last `maxWarnValidityDays` days).

grafik

## logical view

Only warns from the last `maxWarnValidityDays` days are checked.
Every `warnDecayDays` days, `warnDecayAmount` of severity are subtracted from the total severity of the user.
This method does not allow negative severities at any point in time.
The oldest considered warn decides on the amount of severity to subtract.
Hence, all warns that would (together) not increase the severity due to being too old are ignored.

## implementation

The total severity is calculated per warn by considering the severity of all warns after the given warn and subtracting the discount subtrahend corresponding to the given (oldest) warn from the severity.
Then, the maximum discounted severity over all warns is chosen.

See `ModerationService#getTotalSeverityWeight`

## Deploy notes

The config needs to be adapted. Merging this PR alone without any config changes shouldn't change anything about the warn system.

As this is a significant change, it should be communicated properly.

### DIFF
--- a/src/main/java/net/discordjug/javabot/api/routes/user_profile/UserProfileController.java
+++ b/src/main/java/net/discordjug/javabot/api/routes/user_profile/UserProfileController.java
@@ -10,6 +10,7 @@ import net.discordjug.javabot.api.routes.user_profile.model.UserProfileData;
 import net.discordjug.javabot.data.config.BotConfig;
 import net.discordjug.javabot.systems.help.HelpExperienceService;
 import net.discordjug.javabot.systems.help.model.HelpAccount;
+import net.discordjug.javabot.systems.moderation.ModerationService;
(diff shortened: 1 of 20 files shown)


jasonlessenich (approved):

jasonlessenich:
bump

liquiddevelopmentnet (approved):
lgtm

Issue #451, opened 2023-10-23
Add an API to query a member of the Java Discord


gsayson:
**Is your feature request related to a problem? Please describe.**
In a private correspondence between me and @liquiddevelopmentnet , we are discussing the creation of a specific website for voting on specific events, to mitigate voting by bots, and to humanize the process as much as possible.

Since we plan to restrict this to only Discord JUG members, we would need some way of confirming membership.

**Existing APIs and why they don't work**
The current endpoint for looking up users does not provide indication whether a member is in the Discord JUG.

**Describe the solution you'd like**
A new field in the `guilds/{id}/users/{userId}` endpoint, which indicates if the user is member of the guild or not. Some details like time joined may be added.

danthe1st:
Closing that as we decided this wouldn't be needed.

PR #480, opened 2024-02-17
Indicate is user member of guild

Sprokof:
Hi, i added indication for is user member of guild or not


danthe1st:
@liquiddevelopmentnet Is this needed for our purposes?
I assume this refers to #451 which was created but I think we decided to not really need that.

danthe1st:
Hi @Sprokof,
It seems like we are not really needing this feature for our purposes.
Unfortunately, our issues page is not really up-to-date.
If you still want to contribute something to this repository, you could possibly check in the Discord server.

PR #484, opened 2024-03-04
allow users to create threads for their reports

danthe1st:
When reporting a user, this PR adds a button which automatically creates a private thread where the reporter can provide additional information.

image
image
image
image

# Deploy notes
The config `moderationConfig.reportUserThreadHolderId` needs to be set when this is used.

### DIFF
--- a/src/main/java/net/discordjug/javabot/systems/moderation/report/ReportManager.java
+++ b/src/main/java/net/discordjug/javabot/systems/moderation/report/ReportManager.java
@@ -215,8 +215,8 @@ public class ReportManager implements ButtonHandler, ModalHandler {
 		reportEmbed.setDescription("Successfully reported " + "`" + UserUtils.getUserTag(targetUser) + "`!\nYour report has been send to our Moderators.\nIn case you want to supply additional details, please use the \"Create thread\" button below.");
 		hook.sendMessageEmbeds(reportEmbed.build())
 			.addActionRow(Button.secondary(
-					ComponentIdBuilder.build(REPORT_INTERACTION_NAME, "create-thread",reportThread.getId()),
-					"Create thread"))
+					ComponentIdBuilder.build(REPORT_INTERACTION_NAME, "create-thread", reportThread.getId()),
+					"Create thread for providing further details"))
 			.queue();
 	}
 
--- a/src/main/java/net/discordjug/javabot/systems/moderation/report/ReportManager.java
+++ b/src/main/java/net/discordjug/javabot/systems/moderation/report/ReportManager.java
@@ -17,6 +17,7 @@ import net.discordjug.javabot.util.WebhookUtil;
 import net.dv8tion.jda.api.EmbedBuilder;
 import net.dv8tion.jda.api.entities.*;
 import net.dv8tion.jda.api.entities.channel.Channel;
+import net.dv8tion.jda.api.entities.channel.concrete.TextChannel;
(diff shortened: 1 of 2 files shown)


jasonlessenich (approved):

jasonlessenich (approved):

jasonlessenich:
bump

liquiddevelopmentnet:
Which channel in the server do you have in mind for being the thread parent?

danthe1st:
> Which channel in the server do you have in mind for being the thread parent?

I would use the `#general` channel.

src/main/java/net/discordjug/javabot/systems/moderation/report/ReportManager.java
```diff
@@ -147,16 +200,26 @@ protected WebhookMessageCreateAction<Message> handleUserReport(InteractionHook h
...
 		});
 		return null;
 	}
 
+	private void sendReportResponse(InteractionHook hook, User targetUser, EmbedBuilder reportEmbed, ThreadChannel reportThread) {
+		reportEmbed.setDescription("Successfully reported " + "`" + UserUtils.getUserTag(targetUser) + "`!\nYour report has been send to our Moderators.\nIn case you want to supply additional details, please use the \"Create thread\" button below.");
+		hook.sendMessageEmbeds(reportEmbed.build())
+			.addActionRow(Button.secondary(
+					ComponentIdBuilder.build(REPORT_INTERACTION_NAME, "create-thread",reportThread.getId()),
+					"Create thread"))
```
liquiddevelopmentnet, line 219: I'd rename this to something like `Create thread for further discussion` or similar, to make the purpose of this more clear.

liquiddevelopmentnet (approved):

src/main/java/net/discordjug/javabot/systems/moderation/report/ReportManager.java
danthe1st: I don't like using the word `discussion` as that weakly implies that we would generally respond in the threads after handling the report. I rather see these threads as a way for the person reporting to include attachments or other relevant things (e.g. explain a complicated situation better).
I now changed the button to `Create thread for providing further details`.
image

liquiddevelopmentnet (approved):

jasonlessenich:
> > Which channel in the server do you have in mind for being the thread parent?
> 
> I thought about using the `#general` channel.

Wouldn't a forum channel in the staff category be more suitable? Idk if you can restrict a user to only seeing this single thread tho

danthe1st:
> > > Which channel in the server do you have in mind for being the thread parent?
> > 
> > I thought about using the `#general` channel.
> 
> Wouldn't a forum channel in the staff category be more suitable? Idk if you can restrict a user to only seeing this single thread tho

This is not possible AFAIK.
Sending messages in threads requires being able to access thread thread and forum channels provide no way of restricting threads.

Furthermore, I don't think users should be able to view some report channel without ever using that feature.

jasonlessenich:
> > > > Which channel in the server do you have in mind for being the thread parent?
> > > 
> > > I thought about using the `#general` channel.
> > 
> > Wouldn't a forum channel in the staff category be more suitable? Idk if you can restrict a user to only seeing this single thread tho
> 
> This is not possible AFAIK.
> Sending messages in threads requires being able to access thread thread and forum channels provide no way of restricting threads.
> 
> Furthermore, I don't think users should be able to view some report channel without ever using that feature.

Well, they could only have permission to see this channel if they have an open thread 
But yeah, ig #general is fine

PR #488, opened 2024-05-15
use requestor as message link author instead of linked message author

danthe1st:
When sending a message link, the bot automatically responds a copy of the linked message.
The delete button currently allows the author of the original message as well as moderators to delete that copy.

This PR changes that behavior to allow both the author of the message link (to which the bot responds to) and of the author of the original message to delete the message.


danthe1st:
It now allows both users to delete the message.

PR #494, opened 2024-08-24
fix pingable name listener

danthe1st:
This PR changes the `PingableNameListener` (which never really worked before) in the following ways:
- It only changes names of users if the bot has the permission to "interact" with the user (i.e. the bot roles are "above" the user) and not vice-versa.
- Instead of requiring the first 3 characters to be in a alphanumeric range, it requires names to have at least 3 alphanumerical (including underscores) characters.
- Logs are sent to the message log to not spam the moderation log.
- The bot stops adding numbers at the end of generated names.

### DIFF
--- a/src/main/java/net/discordjug/javabot/listener/PingableNameListener.java
+++ b/src/main/java/net/discordjug/javabot/listener/PingableNameListener.java
@@ -15,6 +15,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Random;
 import java.util.Scanner;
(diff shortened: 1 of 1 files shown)


danthe1st:
It seems this is incompatible with disabling member caching in #502.

danthe1st:
To work without member caching (#502), I changed it to use `onGuildMemberUpdate` as `onGuildMemberUpdateNickname` requires the member to be cached.

liquiddevelopmentnet (approved):

PR #502, opened 2024-11-12
native-image compilation, disable member cache

danthe1st:
This PR is in preparation for relocating the bot / reducing the resources required by the bot.

### Changes
- Use GraalVM's [native image](https://www.graalvm.org/latest/reference-manual/native-image/) tool to generate Dockerfiles. This reduces the amount of resources necessary to run the bot.
  - This introduces a closed world assumption meaning all resources and classes used by reflection, JNI or similar mechanisms need to be specified at compile-time. These hints are specified in a class named `RuntimeHintsConfiguration` and a file named `jni-config.json` for JNI hints (necessary for Swing/plotting)
- Disable JDA's member caching as storing member information for 25k members unnecessarily uses up memory.
  - This requires using `retrieve...` instead of `get` operations and `getUser` instead of `getMember` (where Discord doesn't provide member information) in some places
  - I used `complete()` in plotting-related logic but that is executed in the `asyncPool` so that shouldn't cause any issues.
  - Since members are no longer cached, the users with the QOTW champion role cannot be obtained easily any more. The current QOTW champions are now stored in the DB.
- Update Spring (Boot)
- bugfixes
  - The thanking system didn't actually thank the last user for each post because that button isn't disabled at that point
  - fix an issue with the serverlock
  - `/tag-admin edit` had an unnecessary `.deferReply()` that caused exceptions. This has been removed.
  - The metrics updater was not reliably registered
- remove unnecessary DB pooling logs
- Remove the spam links file since that's loaded automatically

### Progress
- [x] explain the changes here (including docs how to build it)
- [x] remove first commit because it's from another PR, squash
- [X] use `-O3`
- [X] properly test functionality
  - [X] all commands including autocomplete
  - [X] other systems like help, automod and QOTW, webhook things, API, etc.
  - [X] ensure other current PRs work
- [X] ensure AJP is working with the API site
- [X] run it as a systemd service
- [X] automatically create backups of data
- [x] actually replace bot keeping the config and DB
  - [x] update config including bot owners and AJP address
- [X] redirect API to new location ~~(via AJP)~~
- [x] run QOTW champion migration
  - [x] ensure current QOTW champions are in DB
- [x] Change Docker Hub repo and update credentials

### Other things noticed:
`/db-admin quick-migrate` doesn't use transactions properly. This can be implemented in a future PR.

### not tested:
Search the web (no token):
If that doesn't work, we can still fix that in the future.


danthe1st:
We aren't using AJP any more.

PR #510, opened 2025-01-13
assign role for pruning

danthe1st:
This PR changes the `/prune` command to assign a role instead of immediately banning users. This allows manual checking in the members tab before banning all of these users.

This PR also adds a few warnings when pruning members that are very likely not intended to be part of the prune.


liquiddevelopmentnet (approved):
lgtm

PR #511, opened 2025-03-09
update dependencies

danthe1st:
This PR updates various dependencies including Spring and H2.

# BEFORE DEPLOYMENT
One of the dependency updates is H2 which uses a different database format now. Before deploying this change, the DB should be exported (`/db-admin export-schema include-data: true`) and then a new DB should be created (e.g. via the DB console using `java -jar ~/.m2/repository/com/h2database/h2/2.3.232/h2-2.3.232.jar`) and the exported script needs to be executed using:
```sql
RUNSCRIPT FROM 'dump.sql'
```

This new DB can then be used instead the old one. If the update needs to be reverted for some reason, I created a backup of the image using the name `dan1st/javabot:backup`. The new version is available under `dan1st/javabot-test:dep-update` (until it's merged)

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -3,10 +3,10 @@ import com.github.jengelman.gradle.plugins.shadow.transformers.*
 
 plugins {
     java
-    id("com.github.johnrengelman.shadow") version "7.1.2"
-    id("org.springframework.boot") version "3.3.5"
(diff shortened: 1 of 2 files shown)


danthe1st:
Deployed successfully.

PR #514, opened 2025-04-13
Fix permission settings for QOTW submissions channel in QOTW jobs

Defective4:
JavaBot is meant to revoke users' permissions to send messages in threads in QOTW submissions channel after the QOTW submission period ends.
Currently after the submission period ends the bot resets all of the channel's role overrides for `@everyone` and doesn't deny the `MESSAGE_SEND` (and other) permissions until the next submission period starts, meaning that all members are allowed to send messages in the submissions channel during that short window of time.
Also since the bot resets the channel permissions at the start and end of submission period and does NOT set other permissions such as `CREATE_PRIVATE_THREADS` and `MESSAGE_ADD_REACTION`, all members are allowed to create new threads manually and add message reactions at all times.
This pull request fixes both of these issues by ensuring that `MESSAGE_SEND` permission is set to denied at the end of the submission period, and by denying the following permissions at both start and end of the period:
- `CREATE_PRIVATE_THREADS`
- `CREATE_PUBLIC_THREADS`
- `MESSAGE_ADD_REACTION`

### DIFF
--- a/src/main/java/net/discordjug/javabot/systems/qotw/jobs/QOTWCloseSubmissionsJob.java
+++ b/src/main/java/net/discordjug/javabot/systems/qotw/jobs/QOTWCloseSubmissionsJob.java
@@ -46,6 +46,7 @@ import java.sql.SQLException;
 import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
+import java.util.Set;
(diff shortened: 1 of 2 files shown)


danthe1st (approved):
LGTM and it works in my tests.

PR #549, opened 2026-06-20
Fix formatting of code longer than Discord's message limit

### LINKED ISSUE #537: Exception formatting code >2k characters
1. **Describe the bug**
When using the "format code" context menu command on messages near or exceeding Discord's 2k character limit, an exception is thrown and the interaction fails.
```
ForkJoinPool.commonPool-worker-219            DIH4JDALogger   WARN   onCommandException was fired, but not handled (No listener registered)
java.lang.IllegalArgumentException: Content may not be longer than 2000 characters! Provided: "[long message here]"
        at net.dv8tion.jda.internal.utils.Checks.check(Checks.java:71)
        at net.dv8tion.jda.internal.utils.Checks.notLonger(Checks.java:175)
        at net.dv8tion.jda.api.utils.messages.AbstractMessageBuilder.setContent(AbstractMessageBuilder.java:66)
        at net.dv8tion.jda.internal.utils.message.AbstractMessageBuilderMixin.setContent(AbstractMessageBuilderMixin.java:45)
        at net.dv8tion.jda.api.interactions.callbacks.IReplyCallback.reply(IReplyCallback.java:238)
        at net.dv8tion.jda.api.interactions.callbacks.IReplyCallback.replyFormat(IReplyCallback.java:486)
        at net.discordjug.javabot.systems.user_commands.format_code.FormatCodeMessageContext.execute(FormatCodeMessageContext.java:28)
        at net.discordjug.javabot.systems.user_commands.format_code.FormatCodeMessageContext.execute(FormatCodeMessageContext.java:16)
        at xyz.dynxsty.dih4jda.InteractionHandler.handleMessageContextCommand(InteractionHandler.java:576)
        at xyz.dynxsty.dih4jda.InteractionHandler.lambda$onMessageContextInteraction$18(InteractionHandler.java:675)
        at java.base@25.0.2/java.util.concurrent.CompletableFuture$AsyncRun.run(CompletableFuture.java:1825)
        at java.base@25.0.2/java.util.concurrent.CompletableFuture$AsyncRun.exec(CompletableFuture.java:1817)
        at java.base@25.0.2/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:511)
        at java.base@25.0.2/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1450)
        at java.base@25.0.2/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:2019)
        at java.base@25.0.2/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:187)
        at org.graalvm.nativeimage.builder/com.oracle.svm.core.thread.PlatformThreads.threadStartRoutine(PlatformThreads.java:832)
        at org.graalvm.nativeimage.builder/com.oracle.svm.core.thread.PlatformThreads.threadStartRoutine(PlatformThreads.java:808)
```

2. **Steps to reproduce**
- Send a message with 1999 characters.
- Right-click the message and select Apps > Format Code

3. **Expected behavior**
The bot should either show an error response, cut off the formatted message or split it in multiple messages.

Neil-Tomar:
## Description of the Changes

The "Format Code" / "Format and Indent Code" message commands and the `/format-code` slash command failed when the target message was longer than Discord's 2000-character limit: the whole snippet was wrapped in a single code block and sent as one message, which Discord rejects — so the command silently did nothing.

This PR splits the formatted output into chunks that each stay under the limit:
- New `Code` class splits content into ≤1980-character chunks (breaking on newlines where possible) and renders each as a code block.
- The commands acknowledge the interaction by replying with the full, unsplit code as an attached file, then post the chunks **in order**.
- New `Language` enum + `Language.fromString(...)` maps the slash command's `format` option to a code-fence tag.

A follow-up PR will add the language-selection dropdown.

### DIFF
--- a/src/main/java/net/discordjug/javabot/systems/user_commands/format_code/Code.java
+++ b/src/main/java/net/discordjug/javabot/systems/user_commands/format_code/Code.java
@@ -18,6 +18,12 @@ public class Code {
 	private Language language;
 	private final String content;
 
+	/**
+	 * Creates a code block for the given language and content.
+	 *
+	 * @param language the language the code is written in, used for syntax highlighting
+	 * @param content  the raw, already-sanitized code to format
+	 */
 	public Code(Language language, String content) {
 		this.language = language;
 		this.content = content;
@@ -31,15 +37,13 @@ public class Code {
 		return language;
 	}
 
-	public void setLanguage(Language language) {
-		this.language = language;
-	}
-
 	/**
 	 * Splits {@link #content} into pieces that each fit within {@link #MAX_SIZE},
 	 * breaking on newlines where possible so lines are not cut in half.
+	 *
+	 * @return the content split into chunks that each fit within the limit
 	 */
-	public List<String> toDiscordChunks() {
+	private List<String> toDiscordChunks() {
 		List<String> chunks = new ArrayList<>();
 		String remaining = content;
 
@@ -59,7 +63,12 @@ public class Code {
 		return chunks;
 	}
 
-	/** Wraps each chunk in a language-tagged Discord code block. */
+	/**
+	 * Splits the content into chunks that each fit within Discord's character limit and wraps
+	 * every chunk in a language-tagged code block.
+	 *
+	 * @return the formatted code-block messages, one per Discord message
+	 */
 	public List<String> toDiscordMessages() {
 		return toDiscordChunks()
 				.stream()
--- a/src/main/java/net/discordjug/javabot/systems/user_commands/format_code/Code.java
+++ b/src/main/java/net/discordjug/javabot/systems/user_commands/format_code/Code.java
@@ -1,0 +1,69 @@
+package net.discordjug.javabot.systems.user_commands.format_code;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Holds a piece of code and its {@link Language}, and turns it into
+ * Discord-friendly representations that respect Discord's 2000-character limit.
+ */
+public class Code {
+
+	/**
+	 * Maximum characters per chunk. Discord's hard limit per message is 2000;
+	 * the remaining headroom covers the surrounding ```language fences.
+	 */
+	private static final int MAX_SIZE = 1980;
+
+	private Language language;
+	private final String content;
+
+	public Code(Language language, String content) {
+		this.language = language;
+		this.content = content;
+	}
+
+	public String getContent() {
+		return content;
+	}
+
+	public Language getLanguage() {
+		return language;
+	}
+
+	public void setLanguage(Language language) {
+		this.language = language;
+	}
+
+	/**
+	 * Splits {@link #content} into pieces that each fit within {@link #MAX_SIZE},
+	 * breaking on newlines where possible so lines are not cut in half.
+	 */
+	public List<String> toDiscordChunks() {
+		List<String> chunks = new ArrayList<>();
+		String remaining = content;
+
+		while (remaining.length() > MAX_SIZE) {
+			int split = remaining.lastIndexOf('\n', MAX_SIZE);
+			if (split <= 0) {
+				// No newline in range (or only at the very start) -> hard cut,
+				// guaranteeing progress so this can never infinite-loop.
+				chunks.add(remaining.substring(0, MAX_SIZE));
+				remaining = remaining.substring(MAX_SIZE);
+			} else {
+				chunks.add(remaining.substring(0, split));
+				remaining = remaining.substring(split + 1); // +1 consumes the '\n'
+			}
+		}
+		chunks.add(remaining);
+		return chunks;
+	}
+
+	/** Wraps each chunk in a language-tagged Discord code block. */
+	public List<String> toDiscordMessages() {
+		return toDiscordChunks()
+				.stream()
+				.map(chunk -> String.format("```%s\n%s\n```", language.getDiscordName(), chunk))
+				.toList();
+	}
+}
--- a/src/main/java/net/discordjug/javabot/systems/user_commands/format_code/Code.java
+++ b/src/main/java/net/discordjug/javabot/systems/user_commands/format_code/Code.java
@@ -15,7 +15,7 @@ public class Code {
 	 */
 	private static final int MAX_SIZE = 1980;
 
-	private Language language;
+	private final Language language;
 	private final String content;
 
 	/**
--- a/src/main/java/net/discordjug/javabot/systems/user_commands/format_code/Code.java
+++ b/src/main/java/net/discordjug/javabot/systems/user_commands/format_code/Code.java
@@ -18,28 +18,34 @@ public class Code {
 	private Language language;
 	private final String content;
 
+	/**
+	 * @param language the language the code is written in, used for syntax highlighting
+	 * @param content  the raw, already-sanitized code to format
+	 */
 	public Code(Language language, String content) {
 		this.language = language;
 		this.content = content;
 	}
 
+	/**
+	 * @return the raw code content
+	 */
 	public String getContent() {
 		return content;
 	}
 
+	/**
+	 * @return the {@link Language} this code is formatted as
+	 */
 	public Language getLanguage() {
 		return language;
-	}
-
-	public void setLanguage(Language language) {
-		this.language = language;
 	}
 
 	/**
 	 * Splits {@link #content} into pieces that each fit within {@link #MAX_SIZE},
 	 * breaking on newlines where possible so lines are not cut in half.
 	 */
-	public List<String> toDiscordChunks() {
+	private List<String> toDiscordChunks() {
 		List<String> chunks = new ArrayList<>();
 		String remaining = content;
 
@@ -59,7 +65,12 @@ public class Code {
 		return chunks;
 	}
 
-	/** Wraps each chunk in a language-tagged Discord code block. */
+	/**
+	 * Splits the content into chunks that each fit within Discord's character limit and wraps
+	 * every chunk in a language-tagged code block.
+	 *
+	 * @return the formatted code-block messages, one per Discord message
+	 */
 	public List<String> toDiscordMessages() {
 		return toDiscordChunks()
 				.stream()
--- a/src/main/java/net/discordjug/javabot/systems/user_commands/format_code/Code.java
+++ b/src/main/java/net/discordjug/javabot/systems/user_commands/format_code/Code.java
@@ -33,10 +33,6 @@ public class Code {
 		return content;
 	}
 
-	public Language getLanguage() {
-		return language;
-	}
-
 	/**
 	 * Splits {@link #content} into pieces that each fit within {@link #MAX_SIZE},
 	 * breaking on newlines where possible so lines are not cut in half.
--- a/src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatAndIndentCodeMessageContext.java
+++ b/src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatAndIndentCodeMessageContext.java
@@ -1,15 +1,19 @@
 package net.discordjug.javabot.systems.user_commands.format_code;
 
-
+import net.discordjug.javabot.util.ExceptionLogger;
 import net.discordjug.javabot.util.IndentationHelper;
+import net.discordjug.javabot.util.Responses;
 import net.discordjug.javabot.util.StringUtils;
+import net.dv8tion.jda.api.entities.channel.middleman.MessageChannel;
 import net.dv8tion.jda.api.events.interaction.command.MessageContextInteractionEvent;
 import net.dv8tion.jda.api.interactions.InteractionContextType;
 import net.dv8tion.jda.api.interactions.commands.build.Commands;
-
+import net.dv8tion.jda.api.utils.FileUpload;
 import org.jetbrains.annotations.NotNull;
 import xyz.dynxsty.dih4jda.interactions.commands.application.ContextCommand;
 
+import javax.annotation.Nonnull;
+import java.nio.charset.StandardCharsets;
 import java.util.List;
 
 /**
@@ -27,9 +31,48 @@ public class FormatAndIndentCodeMessageContext extends ContextCommand.Message {
 
 	@Override
 	public void execute(@NotNull MessageContextInteractionEvent event) {
-		event.replyFormat("```java\n%s\n```", IndentationHelper.formatIndentation(StringUtils.standardSanitizer().compute(event.getTarget().getContentRaw()), IndentationHelper.IndentationType.TABS))
+		String indented = IndentationHelper.formatIndentation(
+				StringUtils.standardSanitizer().compute(event.getTarget().getContentRaw()),
+				IndentationHelper.IndentationType.TABS);
+
+		if (indented.isBlank()) {
+			event.reply("There is no code to format in that message.").setEphemeral(true).queue();
+			return;
+		}
+
+		Code code = new Code(Language.JAVA, indented);
+		List<String> messages = code.toDiscordMessages();
+
+		// Reply with the full code as a file (acknowledges the interaction), then post
+		// the readable code-block chunks in order.
+		FileUpload file = FileUpload.fromData(indented.getBytes(StandardCharsets.UTF_8),
+				"code." + code.getLanguage().getDiscordName());
+		MessageChannel channel = event.getChannel();
+		event.replyFiles(file)
 				.setAllowedMentions(List.of())
-				.setComponents(FormatCodeCommand.buildActionRow(event.getTarget(), event.getUser().getIdLong()))
-				.queue();
+				.queue(
+						success -> sendChunksInOrder(channel, messages, 0, event),
+						error ->  {
+							ExceptionLogger.capture(error, getClass().getSimpleName());
+							Responses.error(event.getHook(), "The message could not be converted into a formatted code block.")
+									.queue();
+						}
+				);
+	}
+
+	private void sendChunksInOrder(MessageChannel channel, List<String> messages, int index, @Nonnull MessageContextInteractionEvent event) {
+		if (index >= messages.size()) {
+			return;
+		}
+		channel.sendMessage(messages.get(index))
+				.setAllowedMentions(List.of())
+				.queue(
+						success -> sendChunksInOrder(channel, messages, index + 1, event),
+						error -> {
+							ExceptionLogger.capture(error, getClass().getSimpleName());
+							Responses.error(event.getHook(), "The message could not be converted into a formatted code block.")
+									.queue();
+						}
+				);
 	}
 }
--- a/src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatAndIndentCodeMessageContext.java
+++ b/src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatAndIndentCodeMessageContext.java
@@ -1,20 +1,13 @@
 package net.discordjug.javabot.systems.user_commands.format_code;
 
-import net.discordjug.javabot.util.ExceptionLogger;
 import net.discordjug.javabot.util.IndentationHelper;
-import net.discordjug.javabot.util.Responses;
 import net.discordjug.javabot.util.StringUtils;
-import net.dv8tion.jda.api.entities.channel.middleman.MessageChannel;
 import net.dv8tion.jda.api.events.interaction.command.MessageContextInteractionEvent;
 import net.dv8tion.jda.api.interactions.InteractionContextType;
 import net.dv8tion.jda.api.interactions.commands.build.Commands;
-import net.dv8tion.jda.api.utils.FileUpload;
 import org.jetbrains.annotations.NotNull;
 import xyz.dynxsty.dih4jda.interactions.commands.application.ContextCommand;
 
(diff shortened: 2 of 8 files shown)


src/main/java/net/discordjug/javabot/systems/user_commands/format_code/Code.java
```diff
@@ -0,0 +1,69 @@
...
+
+	public void setLanguage(Language language) {
+		this.language = language;
+	}
+
+	/**
+	 * Splits {@link #content} into pieces that each fit within {@link #MAX_SIZE},
+	 * breaking on newlines where possible so lines are not cut in half.
+	 */
+	public List<String> toDiscordChunks() {
```
danthe1st: Is there a reason for this being `public`?

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/Language.java
```diff
@@ -0,0 +1,49 @@
...
+
+    /**
+     * Resolves a language from a string (e.g. the value of the /format-code "format"
+     * option) by matching its Discord code-fence name, falling back to {@link #UNKNOWN}.
+     *
+     * @param name the code-fence name to look up (case-insensitive)
+     * @return the matching language, or {@link #UNKNOWN} if none matches
+     */
+    public static Language fromString(String name) {
+        for (Language language : values()) {
```
danthe1st: Instead of this loop, you could use something like this:
```java
try {
     return Language.valueOf(name.toUpperCase());
} catch (IllegalArgumentException e) {
    return UNKNOWN;
}
```

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatAndIndentCodeMessageContext.java
```diff
@@ -27,9 +31,48 @@ public FormatAndIndentCodeMessageContext() {
 
 	@Override
 	public void execute(@NotNull MessageContextInteractionEvent event) {
-		event.replyFormat("```java\n%s\n```", IndentationHelper.formatIndentation(StringUtils.standardSanitizer().compute(event.getTarget().getContentRaw()), IndentationHelper.IndentationType.TABS))
+		String indented = IndentationHelper.formatIndentation(
+				StringUtils.standardSanitizer().compute(event.getTarget().getContentRaw()),
+				IndentationHelper.IndentationType.TABS);
+
+		if (indented.isBlank()) {
+			event.reply("There is no code to format in that message.").setEphemeral(true).queue();
```
danthe1st: If you use `Responses.error(event, "There is no code to format in that message.").queue();`, it would show a nicer error message.
Aside from that, is there any possibility of of the string being blank here?

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatAndIndentCodeMessageContext.java
```diff
@@ -27,9 +31,48 @@ public FormatAndIndentCodeMessageContext() {
...
+				);
+	}
+
+	private void sendChunksInOrder(MessageChannel channel, List<String> messages, int index, @Nonnull MessageContextInteractionEvent event) {
+		if (index >= messages.size()) {
+			return;
+		}
+		channel.sendMessage(messages.get(index))
 				.setAllowedMentions(List.of())
-				.setComponents(FormatCodeCommand.buildActionRow(event.getTarget(), event.getUser().getIdLong()))
```
danthe1st, line 32: The action row including the buttons for deleting and the URL for jumping back are no longer present because you removed them here. Please add them to all messages.

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatAndIndentCodeMessageContext.java
```diff
@@ -27,9 +31,48 @@ public FormatAndIndentCodeMessageContext() {
...
+		// the readable code-block chunks in order.
+		FileUpload file = FileUpload.fromData(indented.getBytes(StandardCharsets.UTF_8),
+				"code." + code.getLanguage().getDiscordName());
+		MessageChannel channel = event.getChannel();
+		event.replyFiles(file)
+				.setAllowedMentions(List.of())
+				.queue(
+						success -> sendChunksInOrder(channel, messages, 0, event),
+						error ->  {
+							ExceptionLogger.capture(error, getClass().getSimpleName());
```
danthe1st: I personally think it isn't necessary to add error logic like this when it "should never happen". If there is no handler, JDA logs the error either way. However, it's fine if you prefer it that way.

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatCodeCommand.java
danthe1st: Since you already introduced a `Language` enum, I think you can iterate over the values there and use that instead of listing them all here. Otherwise, the enum is pretty useless without your follow-up PR.

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatCodeCommand.java
```diff
@@ -92,11 +90,38 @@ public void execute(@NotNull SlashCommandInteractionEvent event) {
...
-							.setAllowedMentions(List.of())
-							.setComponents(buildActionRow(target, event.getUser().getIdLong()))
-							.queue(),
+					target -> sendFormattedCode(event, target, format, indentation),
 					e -> Responses.error(event.getHook(), "Could not retrieve message with id: " + messageId).queue());
 		}
 	}
-}
+
+	private void sendFormattedCode(SlashCommandInteractionEvent event, Message target, String format, String indentation) {
```
danthe1st: There is quite a bit of code duplication between these methods and the ones in the other classes. Please refactor that to only have the common logic once.

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/Code.java
```diff
@@ -0,0 +1,69 @@
...
+
+	private Language language;
+	private final String content;
+
+	public Code(Language language, String content) {
+		this.language = language;
+		this.content = content;
+	}
+
+	public String getContent() {
```
danthe1st, line 32: I think `getContent()` and `setLanguage` are unused. If so, please remove them.

danthe1st (changes requested):
Congratulations on writing your first pull request. I commented on a few things I noticed. Please don't unnecessarily duplicate code.

danthe1st:
Fixes #537

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatAndIndentCodeMessageContext.java
Neil-Tomar: There is a reason for that. The buttons only affect the message they are attached to, so adding them to each message would only allow deletion of that specific message.

Adding buttons to every message would also introduce a visual break inside the code block, which could hurt readability, as shown in the screenshots.

If you'd prefer the buttons to appear only on the last message while still affecting the entire code block, I can look into implementing that. I'm not familiar with that approach yet, but I'm happy to investigate it.

image

Here is with buttons:

image

as you can see it's hard to read through indentation.

danthe1st: If it is a single message, both buttons should definitely be there.
If it consists of multiple messages, at least the `View Original` button should be present in the last message. There could also be a delete button that deletes all codeblock messages on the last message (and the file upload message should also have both buttons).

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/Code.java
```diff
@@ -0,0 +1,78 @@
...
+ */
+public class Code {
+
+	/**
+	 * Maximum characters per chunk. Discord's hard limit per message is 2000;
+	 * the remaining headroom covers the surrounding ```language fences.
+	 */
+	private static final int MAX_SIZE = 1980;
+
+	private Language language;
```
danthe1st: Do you plan on needing to mutate `Language` in the follow-up PR you planned? If not, you could make this `final` or make `Code` a `record`.

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatCodeDispatcher.java
```diff
@@ -0,0 +1,83 @@
...
+import javax.annotation.Nonnull;
+import java.nio.charset.StandardCharsets;
+import java.util.List;
+
+/**
+ * Shared sending logic for the code-formatting commands. Replies with the full code as a
+ * downloadable file, then posts it as one or more ordered code-block messages that each respect
+ * Discord's 2000-character limit.
+ */
+public class FormatCodeDispatcher {
```
danthe1st: This class is only used internally and could be package private.

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatCodeCommand.java
```diff
@@ -56,6 +37,29 @@ public FormatCodeCommand() {
...
+
+	/**
+	 * Builds the action row placed on the file-upload message: a delete button and a "View Original" link.
+	 *
+	 * @param target      the original message linked by the "View Original" button
+	 * @param requesterId the id of the user permitted to delete the message
+	 * @return an action row containing the delete and "View Original" buttons
+	 */
 	@Contract("_ -> new")
 	static @NotNull ActionRow buildActionRow(@NotNull Message target, long requesterId) {
```
danthe1st: I guess this could be moved to `FormatCodeDispatcher` as that's the place for common code but it's fine as-is.

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatCodeDispatcher.java
```diff
@@ -0,0 +1,83 @@
...
+		// The reply both acknowledges the interaction and hands users the full,
+		// un-split code as a downloadable file (so chunking never loses anything).
+		FileUpload file = FileUpload.fromData(
+				code.getContent().getBytes(StandardCharsets.UTF_8),
+				"code." + code.getLanguage().getDiscordName()
+		);
+
+		MessageChannel channel = target.getChannel();
+
+		event.replyFiles(file)
```
danthe1st: What is the reason to reply with a file vs just using `event.reply...` on the first message and skipping the file?

At least if it's only a single chunk, there shouldn't be any file (but I think it's also better without a file with multiple messages).

Make sure to update the Javadoc accordingly.

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatCodeDispatcher.java
```diff
@@ -0,0 +1,83 @@
...
+	 * @param event  the interaction to reply to
+	 * @param target the original message the code came from, used for the channel and the
+	 *               "View Original" / delete buttons
+	 */
+	public static void sendCode(Code code, @Nonnull CommandInteraction event, Message target){
+		if (code.getContent().isBlank()) {
+			Responses.error(event.getHook(), "There is no code to format in that message.").queue();
+			return;
+		}
+		// Currently we always format as Java. A language dropdown will be added in the future.
```
danthe1st: That comment is incorrect. The `Code` can have a language with `/format-code`.

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatCodeDispatcher.java
```diff
@@ -0,0 +1,83 @@
...
+
+	private static void sendChunksInOrder(MessageChannel channel, List<String> messages, int index, Message target, @Nonnull CommandInteraction event) {
+		if (index >= messages.size()) {
+			return;
+		}
+		var action = channel.sendMessage(messages.get(index))
+				.setAllowedMentions(List.of());
+
+		if (index == messages.size() - 1) {
+			action.setComponents(buildActionRow(target, event.getUser().getIdLong()));
```
danthe1st: Please also add the standard delete button _if_ there is only a single message.

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatCodeDispatcher.java
```diff
@@ -0,0 +1,83 @@
...
+				.setComponents(FormatCodeCommand.buildActionRow(target, event.getUser().getIdLong()))
+				.queue(success -> sendChunksInOrder(channel, messages, 0, target,event));
+	}
+
+
+	private static void sendChunksInOrder(MessageChannel channel, List<String> messages, int index, Message target, @Nonnull CommandInteraction event) {
+		if (index >= messages.size()) {
+			return;
+		}
+		var action = channel.sendMessage(messages.get(index))
```
danthe1st, line 61: If you remove the file upload, you can do something like this:
```java
String content = messages.get(index);
MessageRequest<?> action = index == 0 ? event.reply(content) : channel.sendMessage(content);
action.setAllowedMentions(List.of());
```

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/Language.java
```diff
@@ -0,0 +1,117 @@
+package net.discordjug.javabot.systems.user_commands.format_code;
+
+/**
+ * The programming languages supported by the code-formatting commands. Each constant maps a
+ * human-readable {@link #displayName} to the {@link #discordName} tag Discord uses for
+ * syntax highlighting inside code blocks.
+ */
+public enum Language {
+	/**
```
danthe1st, line 9: I think you don't need Javadoc comments for the enum constants here (but please check using `mvn verify`).

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/Language.java
```diff
@@ -0,0 +1,117 @@
...
+	 */
+	TYPESCRIPT("TypeScript", "typescript"),
+	/**
+	 * Extensible Markup Language.
+	 */
+	XML("XML", "xml"),
+	/**
+	 * Fallback used when the language is unrecognised; renders as plain text.
+	 */
+	UNKNOWN("Unknown", "txt");
```
danthe1st: If you are removing the `replyFiles`, I think you can change the `"txt"` to `""` but it's fine if you don't.

danthe1st (changes requested):
These are mostly minor comments. Is there any good reason why users would want a file response in addition to the codeblocks?

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatAndIndentCodeMessageContext.java
danthe1st: If the original message only has attachments or similar, it can be blank.

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/Code.java
danthe1st: `getContent()` is still present.

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatCodeDispatcher.java
Neil-Tomar: I tried dropping the file and replying with the first chunk as suggested, but it hits two problems:

Sending the chunks as plain messages with no reply leaves the interaction unacknowledged → "This application did not respond."
Replying with the first chunk does acknowledge it, but the reply renders as its own attributed block, separate from the follow-up messages, so the code gets split visually mid-block.
image

Replying with the file avoids both: it acknowledges the interaction, and since the reply is a file (not a chunk) the code-block messages stay uniform with no break — plus it's a clean, copyable full version that reads nicely on desktop. The file seemed like the most useful thing to put there.

So, my preference would actually be to keep the file for both the single- and multi-chunk cases — it gives a consistent experience either way. But if you'd rather it was gone, I'm happy to remove it for the single-chunk case (just reply with the one block directly), or if you have a different approach in mind, I'll gladly go with that.

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/Code.java
Neil-Tomar: `getContent()` in now being used in `FormatCodeDispatcher` at line 34 and 44.

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/Language.java
Neil-Tomar: It needs them.

Screenshot 2026-06-24 130656

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatCodeDispatcher.java
danthe1st: In that case, please replace it with an ephemeral response:
```java
Responses.success(event, "Success", "The formatted message is being sent to this channel.").queue();
```
(or similar)

There is no reason to clutter the public channel with the file.

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatCodeDispatcher.java
danthe1st: Resolving as that approach would result in some existing space for users without compact mode.

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/Language.java
```diff
@@ -0,0 +1,117 @@
+package net.discordjug.javabot.systems.user_commands.format_code;
+
+/**
+ * The programming languages supported by the code-formatting commands. Each constant maps a
+ * human-readable {@link #displayName} to the {@link #discordName} tag Discord uses for
+ * syntax highlighting inside code blocks.
+ */
+public enum Language {
```
danthe1st, line 8: If you want to, you can also add JSON there but you don't need to by any means.

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatCodeDispatcher.java
```diff
@@ -0,0 +1,99 @@
...
+	 * @param target the original message the code came from, used for the channel and the
+	 *               "View Original" / delete buttons
+	 */
+	public static void sendCode(Code code, @Nonnull CommandInteraction event, Message target){
+		if (code.getContent().isBlank()) {
+			Responses.error(event, "There is no code to format in that message.").queue();
+			return;
+		}
+
+		List<String> messages = code.toDiscordMessages();
```
danthe1st, line 42: Please show an error if `messages.size() > 5` instead of sending the code.

danthe1st (changes requested):
Overall, your PR looks pretty good.

During testing, I noticed that it's possible to send a single message that can result in a lot of individual (and big). As an example, you can generate the content of such a message with the following code:
```java
System.out.println("{".repeat(30)+"\n"+"a\n".repeat(1000)+"}".repeat(30));
```
Please add a limit to this. I don't want to have to clean that up without knowing the exact number (I think it is possible to trigger more than 500 messages with a single message that way). See the second review comment here for how to fix it.

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatCodeCommand.java
danthe1st: I forgot to mention the following: If you are already at it, could you please also remove the `Collections.reverse(messages);` statement? That shouldn't be there (unrelated to your PR) and results in formatting the wrong message.

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatCodeDispatcher.java
```diff
@@ -21,6 +21,11 @@
  */
 class FormatCodeDispatcher {
 
+	/**
+	 * The maximum number of code-block messages to post inline; longer code is sent only as a file.
```
danthe1st: Please update the Javadoc to say it shows an error in that situation.

danthe1st (changes requested):

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatCodeDispatcher.java
```diff
@@ -22,7 +20,7 @@
 class FormatCodeDispatcher {
 
 	/**
-	 * The maximum number of code-block messages to post inline; longer code is sent only as a file.
+	 * The maximum number of code-block messages to post inline; longer code results as an ERROR.
```
danthe1st: > results as an ERROR.

This is grammatically wrong. Could you please correct it to

> results in an error.

danthe1st (changes requested):
There's one small typo remaining, otherwise it looks good.

danthe1st (approved):
Looks good, I added some minor changes after testing.

PR #550, opened 2026-07-04
feat: delete-all button and language dropdown for multi-message code blocks

Neil-Tomar:
## What
Multi-message formatted code blocks now have two interactions, backed by a small message-linking helper:
- Delete-all button — removes every message of the block in one click
- Language dropdown (ephemeral) — re-colors every chunk to a chosen language
## How
- LinkedMessages resolves the block from the component's message (getHistoryBefore/After), touching only the bot's own messages.
- FormatCodeInteractionHandler handles the delete button and the language dropdown.
- Permission: requester / staff / owner (same rule as the existing delete button).
## Open questions
- The language dropdown is ephemeral per your suggestion, so it's a one-time picker that's gone on reload — happy to make it a permanent public control instead.
- Rare edge case: two simultaneous formats could interleave bot messages; the resolver is best-effort there.

### DIFF
--- a/src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatCodeDispatcher.java
+++ b/src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatCodeDispatcher.java
@@ -6,6 +6,7 @@ import net.dv8tion.jda.api.components.buttons.Button;
 import net.dv8tion.jda.api.entities.Message;
 import net.dv8tion.jda.api.entities.channel.middleman.MessageChannel;
 import net.dv8tion.jda.api.interactions.commands.CommandInteraction;
+import net.dv8tion.jda.api.requests.restaction.MessageCreateAction;
 import org.jetbrains.annotations.Contract;
 import org.jetbrains.annotations.NotNull;
 
@@ -44,7 +45,7 @@ class FormatCodeDispatcher {
 		MessageChannel channel = target.getChannel();
 
 		if (messages.size() > MAX_MESSAGES) {
-			Responses.errorWithTitle(event.getHook(), "Code out of Bound", "The formatted result is too large to send. Please provide a smaller code snippet or use a paste service instead."
+			Responses.errorWithTitle(event.getHook(), "Code out of Bounds", "The formatted result is too large to send. Please provide a smaller code snippet or use a paste service instead."
 			).queue();
 			return;
 		}
@@ -55,26 +56,39 @@ class FormatCodeDispatcher {
 
 	private static void sendChunksInOrder(MessageChannel channel, List<String> messages, int index, Message target, @Nonnull CommandInteraction event, long firstMessageID) {
 		if (index >= messages.size()) {
-			event.getHook().sendMessage("Your message has been set. If needed, you can change the language used for syntax highlighting below.")
-					.setEphemeral(true)
-					.setComponents(FormatCodeInteractionHandler.buildLanguageMenu(event.getUser().getIdLong(), messages.size(), firstMessageID))
-					.queue();
+			event.getHook().deleteOriginal().queue(_ ->
+				event.getHook().sendMessage("Your message has been sent. If needed, you can change the language used for syntax highlighting below.")
+						.setEphemeral(true)
+						.setComponents(FormatCodeInteractionHandler.buildLanguageMenu(event.getUser().getIdLong(), messages.size(), firstMessageID))
+						.queue()
+			);
 			return;
 		}
-		var action = channel.sendMessage(messages.get(index))
+		MessageCreateAction action = channel.sendMessage(messages.get(index))
 				.setAllowedMentions(List.of());
 
 		if (index == messages.size() - 1) {
-			action.setComponents(buildActionRow(target, event.getUser().getIdLong(), messages.size()));
+			if(messages.size() >1) {
+				action.setComponents(buildMultiMessageActionRow(target, event.getUser().getIdLong(), messages.size(), firstMessageID));
+			} else {
+				action.setComponents(buildActionRow(target,event.getUser().getIdLong()));
+			}
 		}
 
 		action.queue(sent -> sendChunksInOrder(channel, messages, index + 1, target, event, index == 0 ? sent.getIdLong() : firstMessageID));
 	}
 
-	@Contract("_,_,_ -> new")
-	static @NotNull ActionRow buildActionRow(@NotNull Message target, long requesterId, int total) {
+	@Contract("_,_,_,_ -> new")
+	static @NotNull ActionRow buildMultiMessageActionRow(@NotNull Message target, long requesterId, int total, long firstMessageID) {
 		return ActionRow.of(
-				FormatCodeInteractionHandler.createDeleteAllButton(requesterId, total),
+				FormatCodeInteractionHandler.createDeleteAllButton(requesterId, total, firstMessageID ),
+				Button.link(target.getJumpUrl(), "View Original"));
+	}
+
+	@Contract("_,_-> new")
+	static @NotNull ActionRow buildActionRow(@NotNull Message target, long requesterId) {
+		return ActionRow.of(
+				InteractionUtils.createDeleteButton(requesterId),
 				Button.link(target.getJumpUrl(), "View Original"));
 	}
 }
--- a/src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatCodeDispatcher.java
+++ b/src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatCodeDispatcher.java
@@ -33,9 +33,9 @@ class FormatCodeDispatcher {
 	 * @param target the original message the code came from, used for the channel and the
 	 *               "View Original" / delete buttons
 	 */
-	public static void sendCode(Code code, @Nonnull CommandInteraction event, Message target){
+	public static void sendCode(Code code, @Nonnull CommandInteraction event, Message target) {
 		if (code.getContent().isBlank()) {
-			Responses.errorWithTitle(event.getHook(), "404 Code not found","There is no code to format in that message.").queue();
+			Responses.errorWithTitle(event.getHook(), "404 Code not found", "There is no code to format in that message.").queue();
 			return;
 		}
 
@@ -44,20 +44,21 @@ class FormatCodeDispatcher {
 		MessageChannel channel = target.getChannel();
 
 		if (messages.size() > MAX_MESSAGES) {
-			Responses.errorWithTitle(event.getHook(), "Output Too Large", "The formatted result is too large to send. Please provide a smaller code snippet or use a paste service instead."
+			Responses.errorWithTitle(event.getHook(), "Code out of Bound", "The formatted result is too large to send. Please provide a smaller code snippet or use a paste service instead."
 			).queue();
 			return;
 		}
 
-		event.getHook().sendMessage("Your message has been formatted. If needed, you can change the language used for syntax highlighting below.")
-				.setEphemeral(true)
-				.setComponents(FormatCodeInteractionHandler.buildLanguageMenu(event.getUser().getIdLong(),messages.size()))
-				.queue(success -> sendChunksInOrder(channel, messages, 0, target,event));
+		sendChunksInOrder(channel, messages, 0, target, event, 0L);
 	}
 
 
-	private static void sendChunksInOrder(MessageChannel channel, List<String> messages, int index, Message target, @Nonnull CommandInteraction event) {
+	private static void sendChunksInOrder(MessageChannel channel, List<String> messages, int index, Message target, @Nonnull CommandInteraction event, long firstMessageID) {
 		if (index >= messages.size()) {
+			event.getHook().sendMessage("Your message has been set. If needed, you can change the language used for syntax highlighting below.")
+					.setEphemeral(true)
+					.setComponents(FormatCodeInteractionHandler.buildLanguageMenu(event.getUser().getIdLong(), messages.size(), firstMessageID))
+					.queue();
 			return;
 		}
 		var action = channel.sendMessage(messages.get(index))
@@ -67,7 +68,7 @@ class FormatCodeDispatcher {
 			action.setComponents(buildActionRow(target, event.getUser().getIdLong(), messages.size()));
 		}
 
-		action.queue(sent -> sendChunksInOrder(channel, messages, index + 1, target, event));
+		action.queue(sent -> sendChunksInOrder(channel, messages, index + 1, target, event, index == 0 ? sent.getIdLong() : firstMessageID));
 	}
 
 	@Contract("_,_,_ -> new")
--- a/src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatCodeDispatcher.java
+++ b/src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatCodeDispatcher.java
@@ -49,7 +49,9 @@ class FormatCodeDispatcher {
 			return;
 		}
 
-		Responses.success(event.getHook(), "Success", "The formatted message is being sent to this channel.")
+		event.getHook().sendMessage("Your message has been formatted. If needed, you can change the language used for syntax highlighting below.")
+				.setEphemeral(true)
+				.setComponents(FormatCodeInteractionHandler.buildLanguageMenu(event.getUser().getIdLong(),messages.size()))
 				.queue(success -> sendChunksInOrder(channel, messages, 0, target,event));
 	}
 
@@ -62,38 +64,16 @@ class FormatCodeDispatcher {
 				.setAllowedMentions(List.of());
 
 		if (index == messages.size() - 1) {
-			if(index == 0){
-				action.setComponents(buildActionRow(target, event.getUser().getIdLong()));
-			} else {
-				action.setComponents(buildActionRow(target));
-			}
+			action.setComponents(buildActionRow(target, event.getUser().getIdLong(), messages.size()));
 		}
 
-		action.queue(success ->
-				sendChunksInOrder(channel, messages, index + 1, target, event));
+		action.queue(sent -> sendChunksInOrder(channel, messages, index + 1, target, event));
 	}
 
-	/**
-	 * Builds the action row placed on the last code-block message.
-	 *
-	 * @param target      the original message linked by the "View Original" button
-	 * @return an action row containing the "View Original" link button
-	 */
-	@Contract("_ -> new")
-	static @NotNull ActionRow buildActionRow(@NotNull Message target) {
-		return ActionRow.of(Button.link(target.getJumpUrl(), "View Original"));
-	}
-
-	/**
-	 * Builds the action row placed on the file-upload message: a delete button and a "View Original" link.
-	 *
-	 * @param target      the original message linked by the "View Original" button
-	 * @param requesterId the id of the user permitted to delete the message
-	 * @return an action row containing the delete and "View Original" buttons
-	 */
-	@Contract("_,_ -> new")
-	static @NotNull ActionRow buildActionRow(@NotNull Message target, long requesterId) {
-		return ActionRow.of(InteractionUtils.createDeleteButton(requesterId),
+	@Contract("_,_,_ -> new")
+	static @NotNull ActionRow buildActionRow(@NotNull Message target, long requesterId, int total) {
+		return ActionRow.of(
+				FormatCodeInteractionHandler.createDeleteAllButton(requesterId, total),
 				Button.link(target.getJumpUrl(), "View Original"));
 	}
 }
--- a/src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatCodeInteractionHandler.java
+++ b/src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatCodeInteractionHandler.java
@@ -112,10 +112,10 @@ public class FormatCodeInteractionHandler implements ButtonHandler, StringSelect
 
 		Member member = event.getMember();
 		if (member == null) {
-			Responses.errorWithTitle(event, "Server Required", "This button may only be used inside a server.").queue();
+			Responses.errorWithTitle(event, "Server Required", "This may only be used inside a server.").queue();
 			return false;
 		}
-		if (!(member.getIdLong() == requesterId || Checks.hasStaffRole(botConfig, member))) {
+		if (member.getIdLong() != requesterId || !Checks.hasStaffRole(botConfig, member)) {
 			Responses.errorWithTitle(event, "Access Denied", "You are not authorized to perform this action.").queue();
 			return false;
 		}
--- a/src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatCodeInteractionHandler.java
+++ b/src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatCodeInteractionHandler.java
@@ -83,7 +83,7 @@ public class FormatCodeInteractionHandler implements ButtonHandler, StringSelect
 
 		LinkedMessages.resolveBefore(event.getMessage(), Integer.parseInt(id[2]),
 				messages -> event.getChannel().purgeMessages(messages),
-				() -> Responses.error(event.getHook(),"Could not delete the code block"));
+				() -> Responses.error(event.getHook(),"Could not delete the code block").queue());
 	}
 
 	@Override
@@ -107,7 +107,7 @@ public class FormatCodeInteractionHandler implements ButtonHandler, StringSelect
 		LinkedMessages.resolveAfter(event.getMessage(), Integer.parseInt(id[2]),
 				messages -> messages.forEach(message ->
 						message.editMessage(withLanguage(message.getContentRaw(), language)).queue()),
-						() -> Responses.error(event.getHook(),"Could not update the code block"));
+						() -> Responses.error(event.getHook(),"Could not update the code block").queue());
 	}
 
 	private boolean canManage(Member member, long requesterId) {
--- a/src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatCodeInteractionHandler.java
+++ b/src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatCodeInteractionHandler.java
@@ -1,0 +1,131 @@
+package net.discordjug.javabot.systems.user_commands.format_code;
+
+import lombok.RequiredArgsConstructor;
+import net.discordjug.javabot.annotations.AutoDetectableComponentHandler;
+import net.discordjug.javabot.data.config.BotConfig;
+import net.discordjug.javabot.util.Checks;
+import net.discordjug.javabot.util.Responses;
+import net.dv8tion.jda.api.components.actionrow.ActionRow;
+import net.dv8tion.jda.api.components.buttons.Button;
+import net.dv8tion.jda.api.components.selections.SelectOption;
+import net.dv8tion.jda.api.components.selections.StringSelectMenu;
+import net.dv8tion.jda.api.entities.Member;
+import net.dv8tion.jda.api.events.interaction.component.ButtonInteractionEvent;
+import net.dv8tion.jda.api.events.interaction.component.StringSelectInteractionEvent;
+import org.jspecify.annotations.NonNull;
+import xyz.dynxsty.dih4jda.interactions.components.ButtonHandler;
+import xyz.dynxsty.dih4jda.interactions.components.StringSelectMenuHandler;
+import xyz.dynxsty.dih4jda.util.ComponentIdBuilder;
+
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * Handles the interactive components on formatted code blocks: the delete-all button and the
+ * language-selection dropdown. Both act on every message of a (possibly multi-message) block,
+ * which is resolved via {@link LinkedMessages}.
+ */
+@AutoDetectableComponentHandler(FormatCodeInteractionHandler.COMPONENT_ID)
+@RequiredArgsConstructor
+public class FormatCodeInteractionHandler implements ButtonHandler, StringSelectMenuHandler {
+	static final String COMPONENT_ID = "format-code-delete";
+	private final BotConfig botConfig;
+
+	/**
+	 * Builds the delete-all button placed on the last message of a code block.
+	 *
+	 * @param requesterID the id of the user allowed to delete the block
+	 * @param total       the number of messages making up the block
+	 * @return the delete-all button
+	 */
+	public static Button createDeleteAllButton(long requesterID, int total){
+		return Button.secondary(ComponentIdBuilder.build(COMPONENT_ID,requesterID,total),"\uD83D\uDDD1️");
+	}
+
+	private static StringSelectMenu languageMenu(String customId) {
+		return StringSelectMenu.create(customId)
+				.setPlaceholder("Change language")
+				.addOptions(Arrays.stream(Language.values())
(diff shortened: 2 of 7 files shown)


src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatCodeInteractionHandler.java
```diff
@@ -0,0 +1,131 @@
...
+
+/**
+ * Handles the interactive components on formatted code blocks: the delete-all button and the
+ * language-selection dropdown. Both act on every message of a (possibly multi-message) block,
+ * which is resolved via {@link LinkedMessages}.
+ */
+@AutoDetectableComponentHandler(FormatCodeInteractionHandler.COMPONENT_ID)
+@RequiredArgsConstructor
+public class FormatCodeInteractionHandler implements ButtonHandler, StringSelectMenuHandler {
+	static final String COMPONENT_ID = "format-code-delete";
```
danthe1st: The select menu isn't about deleting hence the id `format-code-delete` isn't appropriate.

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatCodeInteractionHandler.java
```diff
@@ -0,0 +1,131 @@
...
+
+	/**
+	 * Builds the delete-all button placed on the last message of a code block.
+	 *
+	 * @param requesterID the id of the user allowed to delete the block
+	 * @param total       the number of messages making up the block
+	 * @return the delete-all button
+	 */
+	public static Button createDeleteAllButton(long requesterID, int total){
+		return Button.secondary(ComponentIdBuilder.build(COMPONENT_ID,requesterID,total),"\uD83D\uDDD1️");
```
danthe1st: I am not sure why such a special `1` character is used by `InteractionUtils` but it would be good if you could change it to use a normal `1` instead of it.

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatCodeInteractionHandler.java
```diff
@@ -0,0 +1,131 @@
...
+		}
+
+		event.deferEdit().queue();
+		var channel = event.getChannel();
+		LinkedMessages.resolve(channel, event.getMessage(), Integer.parseInt(id[2]),
+				messages -> messages.forEach(message -> message.delete().queue()));
+	}
+
+	@Override
+		public void handleStringSelectMenu(@NonNull StringSelectInteractionEvent event, @NonNull List<String> values) {
```
danthe1st: The indentation is messed up here.

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatCodeInteractionHandler.java
```diff
@@ -0,0 +1,131 @@
...
+			Responses.error(event, "This button may only be used inside a server.").queue();
+			return;
+		}
+		if (!canManage(member, requesterId)) {
+			Responses.errorWithTitle(event, "Access Denied", "You are not authorized to perform this action.").queue();
+			return;
+		}
+
+		event.deferEdit().queue();
+		var channel = event.getChannel();
```
danthe1st: Could you please use the proper type here? I assume it's a `MessageChannel`?

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatCodeInteractionHandler.java
```diff
@@ -0,0 +1,131 @@
...
+		}
+		if (!canManage(member, requesterId)) {
+			Responses.errorWithTitle(event, "Access Denied", "You are not authorized to perform this action.").queue();
+			return;
+		}
+
+		Language language = Language.fromString(values.getFirst());
+		event.deferEdit().queue();
+
+		LinkedMessages.resolveForward(event.getChannel(), event.getMessage(), Integer.parseInt(id[2]),
```
danthe1st: Is the channel argument really needed here? Can't you derive it from the message?

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/LinkedMessages.java
```diff
@@ -0,0 +1,60 @@
...
+	 */
+	static void resolve(MessageChannel channel, Message triggerMessage, int total, Consumer<List<Message>> onResolved) {
+		if (total <= 1) {
+			onResolved.accept(List.of(triggerMessage));
+			return;
+		}
+		channel.getHistoryBefore(triggerMessage.getIdLong(), total - 1).queue(history -> {
+			List<Message> block = new ArrayList<>(history.getRetrievedHistory());
+			block.add(triggerMessage);
+			onResolved.accept(onlyOwn(channel, block));
```
danthe1st: Please fail (i.e. return an ephemeral error message) _without deleting any messages_ in the following situations:
- The message count after calling `onlyOwn` is different from the expected message count (`total`), Otherwise it would "forget" to delete messages in case there are user messages in between and I think it's better to fail properly rather than performing an incomplete deletion.
- Any of the messages to delete don't start with a codeblock (as a safety mechanism)

You could do that by just adding an "error" `Consumer`.

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/LinkedMessages.java
```diff
@@ -0,0 +1,60 @@
...
+	/**
+	 * Resolves the block ending at {@code triggerMessage} (walking back {@code total} messages) and
+	 * passes the bot's own messages among them to {@code onResolved}.
+	 *
+	 * @param channel        the channel the block is in
+	 * @param triggerMessage the last message of the block (carries the component)
+	 * @param total          the number of messages in the block
+	 * @param onResolved      receives the bot's messages that make up the block
+	 */
+	static void resolve(MessageChannel channel, Message triggerMessage, int total, Consumer<List<Message>> onResolved) {
```
danthe1st: I think you don't need the `channel` argument as you can use `triggerMessage.getChannel()` instead?

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/LinkedMessages.java
```diff
@@ -0,0 +1,60 @@
...
+	/**
+	 * Resolves the block of {@code total} messages sent after {@code anchorMessage} and passes the
+	 * bot's own messages among them to {@code onResolved}.
+	 *
+	 * @param channel       the channel the block is in
+	 * @param anchorMessage the message just before the block (carries the component)
+	 * @param total         the number of messages in the block
+	 * @param onResolved     receives the bot's messages that make up the block
+	 */
+	static void resolveForward(MessageChannel channel, Message anchorMessage, int total, Consumer<List<Message>> onResolved) {
```
danthe1st: I think you don't need the `channel` argument as you can use `anchorMessage.getChannel()` instead?

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatCodeInteractionHandler.java
```diff
@@ -0,0 +1,131 @@
...
+		}
+		if (!canManage(member, requesterId)) {
+			Responses.errorWithTitle(event, "Access Denied", "You are not authorized to perform this action.").queue();
+			return;
+		}
+
+		event.deferEdit().queue();
+		var channel = event.getChannel();
+		LinkedMessages.resolve(channel, event.getMessage(), Integer.parseInt(id[2]),
+				messages -> messages.forEach(message -> message.delete().queue()));
```
danthe1st: Could you please use the `purgeMessages` method here instead? https://javadoc.io/static/net.dv8tion/JDA/5.0.0-alpha.5/net/dv8tion/jda/api/entities/MessageChannel.html#purgeMessages(java.util.List)

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/LinkedMessages.java
```diff
@@ -0,0 +1,60 @@
...
+		channel.getHistoryAfter(anchorMessage.getIdLong(), total).queue(history ->
+				onResolved.accept(onlyOwn(channel, history.getRetrievedHistory())));
```
danthe1st: Please also perform the same message check/error handling approach here as well.

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/LinkedMessages.java
```diff
@@ -0,0 +1,60 @@
...
+	}
+
+	private static List<Message> onlyOwn(MessageChannel channel, List<Message> messages) {
```
danthe1st: I think the `channel` argument isn't used here?

danthe1st (changes requested):

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/LinkedMessages.java
Neil-Tomar: `channel` argument is being used here,

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatCodeInteractionHandler.java
```diff
@@ -80,13 +80,14 @@ public void handleButton(ButtonInteractionEvent event, Button button) {
 		}
 
 		event.deferEdit().queue();
-		var channel = event.getChannel();
-		LinkedMessages.resolve(channel, event.getMessage(), Integer.parseInt(id[2]),
-				messages -> messages.forEach(message -> message.delete().queue()));
+
+		LinkedMessages.resolveBefore(event.getMessage(), Integer.parseInt(id[2]),
+				messages -> event.getChannel().purgeMessages(messages),
+				() -> Responses.error(event.getHook(),"Could not delete the code block"));
```
danthe1st: This is missing a `.queue()`.

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatCodeInteractionHandler.java
```diff
@@ -103,9 +104,10 @@ public void handleStringSelectMenu(@NonNull StringSelectInteractionEvent event,
 		Language language = Language.fromString(values.getFirst());
 		event.deferEdit().queue();
 
-		LinkedMessages.resolveForward(event.getChannel(), event.getMessage(), Integer.parseInt(id[2]),
+		LinkedMessages.resolveAfter(event.getMessage(), Integer.parseInt(id[2]),
 				messages -> messages.forEach(message ->
-						message.editMessage(withLanguage(message.getContentRaw(), language)).queue()));
+						message.editMessage(withLanguage(message.getContentRaw(), language)).queue()),
+						() -> Responses.error(event.getHook(),"Could not update the code block"));
```
danthe1st: This is missing a `.queue()`.

danthe1st (changes requested):

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatCodeInteractionHandler.java
danthe1st: The premise of this comment was false as the `1` was normal. However, there was an invisible character afterwards which is now fixed.

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatCodeInteractionHandler.java
```diff
@@ -0,0 +1,133 @@
...
+
+	/**
+	 * Builds the language-selection dropdown row for a code block.
+	 *
+	 * @param requesterId the id of the user allowed to change the language
+	 * @param total       the number of messages making up the block
+	 * @return an action row containing the language dropdown
+	 */
+	public static ActionRow buildLanguageMenu(long requesterId, int total) {
+		return ActionRow.of(languageMenu(ComponentIdBuilder.build(COMPONENT_ID, requesterId, total)));
```
danthe1st: I noticed it's possible to do the following:
- User A runs the format code command
- User B runs the format code command
- User A deletes the code they formatted
- User A changes the language from their ephemeral message

In this case, the language of the code block of user B is changed. Could you prevent this in one of the following ways?
- In `LinkedMessages.resolveAfter`, check that the last message contains the delete button with the delete button with the correct requestor (= current user) and that no other messages have a delete button
- Change the ephemeral message to be sent _after_ the formatted codeblocks, use `resolveBefore` and add an additional argument containing the message ID of the first code block message to the ID here. If this doesn't match, it is a failure.

I think the first approach is easier to do so I recommend that one.

danthe1st (changes requested):

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatCodeInteractionHandler.java
```diff
@@ -105,9 +110,46 @@ public void handleStringSelectMenu(@NonNull StringSelectInteractionEvent event,
...
+					if (ownsBlock(messages, event.getUser().getIdLong())) {
+						messages.forEach(message ->
+								message.editMessage(withLanguage(message.getContentRaw(), language)).queue());
+					} else {
+						Responses.error(event.getHook(),"Could not update the code block — it may have changed or no longer be yours.").queue();
+					}
+				}, () -> Responses.error(event.getHook(), "Could not update the code block — it may have changed or no longer be yours.").queue());
+	}
+
+	private static boolean ownsBlock(List<Message> messages, long userId) {
```
danthe1st: Could you please change this name to something more descriptive like `isCodeBlockMessagesOwnedBy` or `isFormattedCodeMessagesOwnedBy`?

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatCodeInteractionHandler.java
```diff
@@ -105,9 +110,46 @@ public void handleStringSelectMenu(@NonNull StringSelectInteractionEvent event,
...
+		Message last = messages.getFirst();
+		boolean lastIsUsers = buttonsOf(last).stream().anyMatch(button -> isDeleteButtonOf(button, userId));
+		boolean noOtherHasButton = messages.size() <= 1 ;
+		for (Message message : messages.subList(0,messages.size()-1)){
+			for(Button button: buttonsOf(message)){
```
danthe1st: Formatted code block messages shouldn't have any componens except for the last message. I think the following would be easier:
```java
for (int i = 1; i < messages.size(); i++) {
    if (!message.getComponents().isEmpty()) {
        return false;
    }
}
return messages.size() >= 1 && hasDeleteButtonOf(messages.getFirst(), userId)
```

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatCodeInteractionHandler.java
```diff
@@ -105,9 +110,46 @@ public void handleStringSelectMenu(@NonNull StringSelectInteractionEvent event,
...
+				if(isDeleteButton(button)){
+					noOtherHasButton = true;
```
danthe1st: Here you are basically requiring that there is a delete button but you are excluding the chronologically _first_ message from the list using `subList` which makes no sense.

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatCodeInteractionHandler.java
```diff
@@ -105,9 +110,46 @@ public void handleStringSelectMenu(@NonNull StringSelectInteractionEvent event,
 		event.deferEdit().queue();
 
 		LinkedMessages.resolveAfter(event.getMessage(), Integer.parseInt(id[2]),
```
danthe1st: Could you please document that `resolveAfter` and `resolveBefore` return the messages ordered [from newest to oldest](https://javadoc.io/static/net.dv8tion/JDA/6.5.0/net/dv8tion/jda/api/entities/MessageHistory.html#getRetrievedHistory())?

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatCodeInteractionHandler.java
```diff
@@ -105,9 +110,46 @@ public void handleStringSelectMenu(@NonNull StringSelectInteractionEvent event,
...
+			for(Button button: buttonsOf(message)){
+				if(isDeleteButton(button)){
+					noOtherHasButton = true;
+				}
+			}
+		}
+		return lastIsUsers && noOtherHasButton;
+	}
+
+	private static List<Button> buttonsOf(Message message) {
```
danthe1st: I don't think you'll need a `buttonsOf` method, you can just create a method like `hasDeleteButtonOf` that checks whether there is a delete button of a given user. Such a method could then just iterate over the components/buttons and checks for the delete button.

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatCodeInteractionHandler.java
danthe1st: If this method requires a certain order of the messages, please document this in a Javadoc.

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatCodeInteractionHandler.java
```diff
@@ -105,9 +110,46 @@ public void handleStringSelectMenu(@NonNull StringSelectInteractionEvent event,
...
+		}
+		return new ArrayList<>();
+	}
+
+	private static boolean isDeleteButton(Button button) {
+		return button.getCustomId() != null && button.getCustomId().startsWith(COMPONENT_ID + ":");
+	}
+
+	private static boolean isDeleteButtonOf(Button button, long userId) {
+		return isDeleteButton(button)
```
danthe1st: I think it would be easier to use `ComponentIdBuilder.split` and:
- verify that there are sufficiently many elements
- verify that the element at index `1` equals `COMPONENT_ID`
- verify that the element at index `1` equals `String.valueOf(userId)`

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatCodeInteractionHandler.java
```diff
@@ -105,9 +110,46 @@ public void handleStringSelectMenu(@NonNull StringSelectInteractionEvent event,
...
 		LinkedMessages.resolveAfter(event.getMessage(), Integer.parseInt(id[2]),
-				messages -> messages.forEach(message ->
-						message.editMessage(withLanguage(message.getContentRaw(), language)).queue()),
-						() -> Responses.error(event.getHook(),"Could not update the code block").queue());
+				messages -> {
+					if (ownsBlock(messages, event.getUser().getIdLong())) {
+						messages.forEach(message ->
+								message.editMessage(withLanguage(message.getContentRaw(), language)).queue());
+					} else {
+						Responses.error(event.getHook(),"Could not update the code block — it may have changed or no longer be yours.").queue();
```
danthe1st: "no longer be yours" is a bit confusing in my opinion. What about "The code block could not be updated. The messages may have been deleted."

Also, I assume you are not able to delete it because the original message is ephemeral? Is this correct (this is perfectly fine)?

danthe1st (changes requested):

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatCodeDispatcher.java
```diff
@@ -44,56 +44,37 @@ public static void sendCode(Code code, @Nonnull CommandInteraction event, Messag
 		MessageChannel channel = target.getChannel();
 
 		if (messages.size() > MAX_MESSAGES) {
-			Responses.errorWithTitle(event.getHook(), "Output Too Large", "The formatted result is too large to send. Please provide a smaller code snippet or use a paste service instead."
+			Responses.errorWithTitle(event.getHook(), "Code out of Bound", "The formatted result is too large to send. Please provide a smaller code snippet or use a paste service instead."
```
danthe1st: I think you meant to say `Code Out of Bounds` (plural).

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatCodeDispatcher.java
```diff
@@ -44,56 +44,37 @@ public static void sendCode(Code code, @Nonnull CommandInteraction event, Messag
...
-		Responses.success(event.getHook(), "Success", "The formatted message is being sent to this channel.")
-				.queue(success -> sendChunksInOrder(channel, messages, 0, target,event));
+		sendChunksInOrder(channel, messages, 0, target, event, 0L);
 	}
 
 
-	private static void sendChunksInOrder(MessageChannel channel, List<String> messages, int index, Message target, @Nonnull CommandInteraction event) {
+	private static void sendChunksInOrder(MessageChannel channel, List<String> messages, int index, Message target, @Nonnull CommandInteraction event, long firstMessageID) {
 		if (index >= messages.size()) {
+			event.getHook().sendMessage("Your message has been set. If needed, you can change the language used for syntax highlighting below.")
```
danthe1st: `sent`, not `set`.

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatCodeDispatcher.java
```diff
@@ -44,56 +44,37 @@ public static void sendCode(Code code, @Nonnull CommandInteraction event, Messag
...
+					.setEphemeral(true)
+					.setComponents(FormatCodeInteractionHandler.buildLanguageMenu(event.getUser().getIdLong(), messages.size(), firstMessageID))
+					.queue();
 			return;
 		}
 		var action = channel.sendMessage(messages.get(index))
```
danthe1st: Would it be possible to specify the type explicitly here instead of using `var`?

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatCodeDispatcher.java
```diff
@@ -44,56 +44,37 @@ public static void sendCode(Code code, @Nonnull CommandInteraction event, Messag
...
-	 * @param requesterId the id of the user permitted to delete the message
-	 * @return an action row containing the delete and "View Original" buttons
-	 */
-	@Contract("_,_ -> new")
-	static @NotNull ActionRow buildActionRow(@NotNull Message target, long requesterId) {
-		return ActionRow.of(InteractionUtils.createDeleteButton(requesterId),
+	@Contract("_,_,_ -> new")
+	static @NotNull ActionRow buildActionRow(@NotNull Message target, long requesterId, int total) {
+		return ActionRow.of(
+				FormatCodeInteractionHandler.createDeleteAllButton(requesterId, total),
```
danthe1st: I think it would make sense to add the first message ID to the delete button and verify that it is actually the first message before deleting. I think that should allow you using (almost) the same logic for finding and checking the messages.

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatCodeInteractionHandler.java
```diff
@@ -0,0 +1,136 @@
...
+	 * Builds the delete-all button placed on the last message of a code block.
+	 *
+	 * @param requesterID the id of the user allowed to delete the block
+	 * @param total       the number of messages making up the block
+	 * @return the delete-all button
+	 */
+	public static Button createDeleteAllButton(long requesterID, int total) {
+		return Button.secondary(ComponentIdBuilder.build(COMPONENT_ID, requesterID, total), "\uD83D\uDDD1\uFE0F");
+	}
+
```
danthe1st, line 46: Could you please move the delete logic together and the language change logic together? e.g. first the methods for creating and handling the delete button and then the methods for creating and handling the language menu.

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatCodeInteractionHandler.java
```diff
@@ -0,0 +1,136 @@
...
+		LinkedMessages.resolveAfter(event.getMessage(), Integer.parseInt(id[2]), messages -> {
+			if (messages.getLast().getIdLong() == Long.parseLong(id[3])) {
+				messages.forEach(message -> message.editMessage(withLanguage(message.getContentRaw(), language)).queue());
+			} else {
+				Responses.errorWithTitle(event.getHook(), "Merge Conflict", "The code block could not be updated. The messages may have been deleted.").queue();
+			}
+		}, () -> Responses.errorWithTitle(event.getHook(), "Undefined Behavior", "Could not update the code block").queue());
+	}
+
+	private boolean canManage(Member member, long requesterId) {
```
danthe1st: Could you please rename this method `isMemberAuthorizedToDeleteMessagesOf`? Also, I think the `isOwner` check is redundant (but I don't have a problem with you keeping this).

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatCodeInteractionHandler.java
```diff
@@ -0,0 +1,136 @@
...
+		if (!canManage(member, requesterId)) {
+			Responses.errorWithTitle(event, "Access Denied", "You are not authorized to perform this action.").queue();
+			return;
+		}
+
+		event.deferEdit().queue();
+
+		LinkedMessages.resolveBefore(event.getMessage(), Integer.parseInt(id[2]),
+				messages -> event.getChannel().purgeMessages(messages),
+				() -> Responses.errorWithTitle(event.getHook(), "Undefined Behavior", "Could not delete the code block").queue());
```
danthe1st: I don't think the "Undefined Behavior" adds much of value. I think it's better to not use a title here.

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatCodeInteractionHandler.java
```diff
@@ -0,0 +1,136 @@
...
+		}
+
+		Language language = Language.fromString(values.getFirst());
+		event.deferEdit().queue();
+
+		LinkedMessages.resolveAfter(event.getMessage(), Integer.parseInt(id[2]), messages -> {
+			if (messages.getLast().getIdLong() == Long.parseLong(id[3])) {
+				messages.forEach(message -> message.editMessage(withLanguage(message.getContentRaw(), language)).queue());
+			} else {
+				Responses.errorWithTitle(event.getHook(), "Merge Conflict", "The code block could not be updated. The messages may have been deleted.").queue();
```
danthe1st: This isn't really a merge conflict. If the titles don't add much of value, please don't specify them.

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatCodeInteractionHandler.java
```diff
@@ -0,0 +1,136 @@
...
+		}
+		if (!canManage(member, requesterId)) {
+			Responses.errorWithTitle(event, "Access Denied", "You are not authorized to perform this action.").queue();
+			return;
+		}
+
+		Language language = Language.fromString(values.getFirst());
+		event.deferEdit().queue();
+
+		LinkedMessages.resolveAfter(event.getMessage(), Integer.parseInt(id[2]), messages -> {
```
danthe1st: Are you sure `resolveAfter` is still correct when you are sending the ephemeral message after the codeblocks?

danthe1st (changes requested):

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatCodeDispatcher.java
danthe1st: If you want to send the ephemeral response after the code block messages, I think you can do the following:
```java
// use deleteOriginal() to delete the intermediate message sent by deferReply(false)
event.getHook().deleteOriginal().queue(_ -> 
    event.getHook().sendMessage("Your message has been sent. If needed, you can change the language used for syntax highlighting below.")
    ...
    .queue();
);
```

Otherwise you would need to check the last message and not the first message I think (the goal is to check the message that is furthest from the anchor).

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatCodeInteractionHandler.java
danthe1st: Note: This is assuming you are changing the interaction to use `deleteOriginal()` to move the message below.

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatCodeDispatcher.java
Neil-Tomar: There's one case this can't cover: when the code block is a single message. Discord only assigns a message's ID once the message is sent, but the button has to be attached to the message before it's sent — so for a single-message block, that message's own ID isn't available at the point its button is built. For multi-message blocks this isn't an issue, since the button sits on the last chunk and the first chunk has already been sent by then.

If you want, I can skip the first-message check for the single-message case.

danthe1st: I think you can just use `InteractionUtils.createDeleteButton()` for the single message case.

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatCodeInteractionHandler.java
```diff
@@ -65,58 +76,51 @@ public static ActionRow buildLanguageMenu(long requesterId, int total, long firs
...
+
+	private boolean isValid(ComponentInteraction event) {
 		String[] id = ComponentIdBuilder.split(event.getComponentId());
 		long requesterId = Long.parseLong(id[1]);
 
 		Member member = event.getMember();
 		if (member == null) {
-			Responses.errorWithTitle(event, "Server Required", "This menu may only be used inside a server.").queue();
-			return;
+			Responses.errorWithTitle(event, "Server Required", "This button may only be used inside a server.").queue();
```
danthe1st: Since I think `isValid` is also used for the menu which isn't a button, I think `This may only be used inside a server.` is better.

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatCodeInteractionHandler.java
```diff
@@ -65,58 +76,51 @@ public static ActionRow buildLanguageMenu(long requesterId, int total, long firs
...
+			return false;
 		}
-		if (!canManage(member, requesterId)) {
+		if (!(member.getIdLong() == requesterId || Checks.hasStaffRole(botConfig, member))) {
```
danthe1st: Could you please change this to `if (member.getIdLong() != requesterId || !Checks.hasStaffRole(botConfig, member))`? (basically just De Morgan) While it is doing the same thing, I think it is easier to read.

danthe1st (changes requested):
Two very minor things. If you fixed those, I'll start testing everything.

src/main/java/net/discordjug/javabot/systems/user_commands/format_code/FormatCodeInteractionHandler.java
danthe1st: My fault, this should be `&&`. I forgot to update it :)

danthe1st (approved):
LGTM

PR #551, opened 2026-07-11
log thread/forum post deletions

danthe1st:
This PR logs forum posts/thread deletions as if they were messages.
When a forum post is deleted, the message content of the "start message" is logged.
In case of normal threads, this just logs the name of the thread.

### DIFF
--- a/src/main/java/net/discordjug/javabot/data/h2db/message_cache/MessageCacheListener.java
+++ b/src/main/java/net/discordjug/javabot/data/h2db/message_cache/MessageCacheListener.java
@@ -69,7 +69,6 @@ public class MessageCacheListener extends ListenerAdapter {
 			messageCache.cache.remove(message);
 		});
 	}
-	
 
 	/**
 	 * Checks whether the given message should be ignored by the cache.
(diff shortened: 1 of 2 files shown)


src/main/java/net/discordjug/javabot/data/h2db/message_cache/MessageCacheListener.java
```diff
@@ -50,13 +52,24 @@ public void onMessageUpdate(@NotNull MessageUpdateEvent event) {
...
+	private void processMessageDeletion(long messageId, GuildChannel channel) {
+		Optional<CachedMessage> optional = messageCache.cache.stream().filter(m -> m.getMessageId() == messageId).findFirst();
 		optional.ifPresent(message -> {
-			messageCache.sendDeletedMessageToLog(event.getGuild(), event.getChannel(), message);
+			messageCache.sendDeletedMessageToLog(channel.getGuild(), channel, message);
 			messageCache.cache.remove(message);
 		});
 	}
-
+
```
Neil-Tomar: You could remove the trailing whitespace on this line, but it's fine either way.

danthe1st: Thanks. I removed that line since having two empty lines directly after each other is redundant either way.

liquiddevelopmentnet (approved):

PR #553, opened 2026-07-17
Feat/cross channel spam automod

Neil-Tomar:
## Summary

Adds a second spam automod — `checkCrossChannelSpam` in `AutoMod` — that catches a pattern the existing same-channel spam check misses: a single user rapidly spreading messages across several different channels.

## What it does

- Flags a user who posts in more than a configured number of **distinct** channels within a configured time window.
- On a match, applies the existing 6-hour spam timeout and deletes the matched messages.
- Both thresholds live in `ModerationConfig`, so they're tunable per guild.

## How it works

- **`CachedMessage` now stores a `channelId`**, taken from the live message when it's cached — so the cache knows which channel each message came from. That's what lets the check count distinct channels **without any extra Discord API calls**; only the timeout and the deletions talk to Discord.
- **`checkCrossChannelSpam`** reads the author's recent messages from the `MessageCache` within the window, gathers the distinct channels, and passes the messages to `handleSpam` once the count crosses the threshold.
- **`handleSpam`** gains a `List<CachedMessage>` overload that times the user out and deletes each message by id.
- New `ModerationConfig` fields: `crossChannelSpamWindowSeconds` and `crossChannelSpamMinChannels`.

## Notes

- `channelId` isn't persisted to the database yet — it defaults to `-1` when read from the DB, per your suggestion. If it holds up, I can follow with a migration to store it properly.
- The trigger currently uses the **distinct-channel count only**. Your spec also mentioned a message count, so I'm happy to add a separate minimum-messages threshold if you'd like both.


src/main/java/net/discordjug/javabot/data/h2db/message_cache/dao/MessageCacheRepository.java
```diff
@@ -119,6 +119,7 @@ private CachedMessage read(ResultSet rs) throws SQLException {
 		return new CachedMessage(
 				rs.getLong("message_cache.message_id"),
 				rs.getLong("author_id"),
+				rs.getLong("channel_id"),
```
danthe1st, line 123: You are trying to get the channel ID from the database here even though that field doesn't exist there. Please either change that to `-1` or create a database migration.

_If_ you want to create a database migration, please do the following:
- Create the SQL script for adding the column in https://github.com/Java-Discord/JavaBot/tree/main/src/main/resources/database/migrations
- You can run that script with `/db-admin migrate <script name>` (you might need to temporarily remove [this line of code](https://github.com/Java-Discord/JavaBot/blob/main/src/main/java/net/discordjug/javabot/data/h2db/commands/MigrateSubcommand.java#L96) but please don't commit that change).
- Then you can change the logic for inserting and retrieving.

src/main/java/net/discordjug/javabot/systems/moderation/AutoMod.java
```diff
@@ -108,18 +111,33 @@ private void checkNewMessageAutomod(@Nonnull Message message) {
...
 			handleSpam(message, message.getMember());
 		}
 
 		checkContentAutomod(message);
+		checkCrossChannelSpam(message);
+	}
+
+	private void checkCrossChannelSpam(@Nonnull Message message) {
+		List<CachedMessage> spamMessages = new ArrayList<>();
+		messageCache.getMessagesAfter(message.getTimeCreated().minusSeconds((botConfig.get(message.getGuild()).getModerationConfig()).getCrossChannelSpamWindowSeconds())).stream()
```
danthe1st: I think you forgot to add a check to ensure this is only executed if the window is strictly positive/nonzero?
```java
int spamWindowSeconds = botConfig.get(message.getGuild()).getModerationConfig()).getCrossChannelSpamWindowSeconds();
if (spamWindowSeconds <= 0) {
    return;
}
messageCache.getAfter(message.getTimeCreated().minusSeconds(spamWindowSeconds))
    ...
```

src/main/java/net/discordjug/javabot/systems/moderation/AutoMod.java
```diff
@@ -108,18 +111,33 @@ private void checkNewMessageAutomod(@Nonnull Message message) {
...
+				.filter(cachedMessage -> cachedMessage.getAuthorId() == message.getAuthor().getIdLong())
```
danthe1st: Please add a check that prevents the new message being included in the count (`message.getIdLong() != cachedMessage.getMessageId())`.

src/main/java/net/discordjug/javabot/systems/moderation/AutoMod.java
```diff
@@ -108,18 +111,33 @@ private void checkNewMessageAutomod(@Nonnull Message message) {
...
 		if (spamCount >= 5) {
 			handleSpam(message, message.getMember());
 		}
 
 		checkContentAutomod(message);
+		checkCrossChannelSpam(message);
+	}
+
+	private void checkCrossChannelSpam(@Nonnull Message message) {
+		List<CachedMessage> spamMessages = new ArrayList<>();
```
danthe1st, line 128: If a user spams across multiple channels but sends two messages in the same channel, this would only delete one message in that channel.
Alternative approach:
```java
List<CachedMessage> spamMessages = new ArrayList<>();
Set<Long> channels = new HashSet<>();

int spamWindowSeconds = botConfig.get(message.getGuild()).getModerationConfig()).getCrossChannelSpamWindowSeconds();
if (spamWindowSeconds <= 0) { // see other comment
    return;
}
for (CachedMessage cachedMessage : messageCache.getAfter(message.getTimeCreated().minusSeconds(spamWindowSeconds))) {
    if (cachedMessage.getAuthorId() == message.getAuthor().getIdLong() && message.getIdLong() != cachedMessage.getMessageId())) {
        channels.add(cachedMessage.getChannelId());
        spamMessages.add(cachedMessage);
    }
}
if(spamMessages.size() >= botConfig.get(message.getGuild()).getModerationConfig().getCrossChannelSpamMinChannels()){
    handleSpam(spamMessages,message);
}

File: src/main/java/net/discordjug/javabot/systems/moderation/AutoMod.java [hunk 5]
```diff
@@ -174,6 +192,22 @@ private void handleSpam(@Nonnull Message msg, Member member) {
 		msg.delete().queue();
 	}
 
+	private void handleSpam(@Nonnull List<CachedMessage> cachedMessages, Message message) {
+		Guild guild = message.getGuild();
+		moderationService
```
danthe1st, line 200: While creating an overload of `handleSpam` is fine, please create one method that performs the timeout and don't duplicate code.

You could also just have one `handleSpam` method that doesn't do deletion:
```java
private void handleSpam(User author, MessageChannel channel) {
    moderationService.timeout(
						author,
						"Automod: Spam",
						channel.getGuild().getSelfMember(),
						Duration.of(6, ChronoUnit.HOURS),
						channel,
						false
				);
}
```

The deletion could then be done by the callers (which are different).

There is no reason to duplicate code here.

src/main/java/net/discordjug/javabot/systems/moderation/AutoMod.java
```diff
@@ -218,7 +252,7 @@ public boolean hasSuspiciousLink(@NotNull Message message) {
 	 * Checks whether the given message contains a discord invite link.
 	 *
 	 * @param message The Message to check.
-	 * @return True if an invite is found and False if not.
+	 * @return True if an invitation is found and False if not.
```
danthe1st: Server invites are generally called invites. Calling them invitation doesn't improve this.

danthe1st (changes requested):

src/main/java/net/discordjug/javabot/systems/moderation/AutoMod.java
```diff
@@ -176,24 +183,20 @@ private void doAutomodActions(Message message, String reason) {
...
 	}
 
 	private void handleSpam(@Nonnull List<CachedMessage> cachedMessages, Message message) {
-		Guild guild = message.getGuild();
+		markSpam(message);
+		cachedMessages.forEach(cachedMessage -> message.getGuild().getTextChannelById(cachedMessage.getChannelId())
+				.deleteMessageById(cachedMessage.getMessageId()).queue());
+	}
+
+	private void markSpam(@Nonnull Message message) {
```
danthe1st: Could you please rename this to `timeoutForSpam` (or similar) because it isn't really marking anything for spam.

src/main/java/net/discordjug/javabot/systems/moderation/AutoMod.java
```diff
@@ -252,7 +252,7 @@ public boolean hasSuspiciousLink(@NotNull Message message) {
 	 * Checks whether the given message contains a discord invite link.
 	 *
 	 * @param message The Message to check.
-	 * @return True if an invitation is found and False if not.
+	 * @return True if an invites is found and False if not.
```
danthe1st: "An invite" is singular. I have no idea why you tried to change that.

src/main/resources/database/migrations/07-19-2026_add_channel_column_to_cache.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE message_cache ADD COLUMN channel_id BIGINT DEFAULT -1;
```
danthe1st: Since you decided to create a DB table, I think it makes sense to also populate it in `MessageCacheRepository#insertList`.

danthe1st (changes requested):

src/main/resources/database/migrations/07-19-2026_add_channel_column_to_cache.sql
danthe1st: Note: When you do, please try to not (unnecessarily) modify anything outside of `insertList` (and also don't reformat anything) to prevent merge conflicts.

src/main/java/net/discordjug/javabot/systems/moderation/AutoMod.java
Neil-Tomar: It was recommended by my IDE.

danthe1st (approved):
LGTM, thanks for the PR

danthe1st:
Deployment Notes:
The database migration `07-19-2026_add_channel_column_to_cache.sql` should be executed right after deploying. Once that is done, the bot should be restarted once more _without_ running `/redeploy` to ensure the message cache is loaded correctly.

PR #557, opened 2026-07-23
move database migrations to schema

danthe1st:
This PR moves all database migrations to the schema.sql (provided they aren't there already) and deletes them as having all database changes run on the first setup simplifies initially getting started.

This also adds some sanity tests to ensure the schema.sql can be loaded properly.


liquiddevelopmentnet (approved):
lgtm!

PR #558, opened 2026-08-08
Update ban message when user is unbanned

danthe1st:
There is no good way to notify users about being unbanned (since we don't want to collect personal data like E-Mail addresses). This PR updates the ban message (which is automatically sent to users when they are banned) when the user is unbanned.

### DIFF
--- a/src/main/java/net/discordjug/javabot/systems/moderation/ModerationService.java
+++ b/src/main/java/net/discordjug/javabot/systems/moderation/ModerationService.java
@@ -10,6 +10,7 @@ import net.discordjug.javabot.util.ExceptionLogger;
 import net.discordjug.javabot.util.Responses;
 import net.discordjug.javabot.util.UserUtils;
 import net.dv8tion.jda.api.EmbedBuilder;
+import net.dv8tion.jda.api.JDA;
 import net.dv8tion.jda.api.entities.*;
 import net.dv8tion.jda.api.entities.channel.middleman.MessageChannel;
 import net.dv8tion.jda.api.utils.MarkdownUtil;
@@ -20,16 +21,20 @@ import org.springframework.stereotype.Service;
 
 import lombok.RequiredArgsConstructor;
 
+import javax.annotation.CheckReturnValue;
 import javax.annotation.Nonnull;
 import java.time.Duration;
 import java.time.Instant;
 import java.time.LocalDateTime;
 import java.time.ZoneOffset;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.TimeUnit;
+import java.util.function.Consumer;
 
 /**
  * This service provides methods for performing moderation actions, like banning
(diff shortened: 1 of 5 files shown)


src/main/java/net/discordjug/javabot/systems/moderation/UnbanCommand.java
```diff
@@ -43,10 +43,13 @@ protected ReplyCallbackAction handleModerationCommand(@NotNull SlashCommandInter
...
-		if (moderationService.unban(id, reasonOption.getAsString(), event.getMember(), event.getChannel(), quiet)) {
-			return Responses.success(event, "User Unbanned", "User with id `%s` has been unbanned.", id);
-		} else {
-			return Responses.warning(event, "Could not find banned User with id `%s`", id);
-		}
+		moderationService.unban(id, reasonOption.getAsString(), event.getMember(), event.getChannel(), quiet)
+			.thenAccept(success -> {
+				Responses.success(event.getHook(), "User Unbanned", "User with id `%s` has been unbanned.", id).queue();
+			}).exceptionally(failed -> {
+				Responses.warning(event, "Could not find banned User with id `%s` or a different error occured: `%s`", id, failed.getMessage()).queue();
```
liquiddevelopmentnet: shouldn't this be `event.getHook()`? the interaction is acked before the warning fires no?

src/main/java/net/discordjug/javabot/systems/moderation/ModerationService.java
```diff
@@ -281,23 +286,44 @@ private void banAndSendGuildNotifications(User user, String reason, Member banne
...
 
-	private boolean isBanned(@NotNull Guild guild, long userId) {
-		return guild.retrieveBanList().complete()
-				.stream().map(Guild.Ban::getUser)
-				.map(User::getIdLong).toList().contains(userId);
+	private void notifyUserAboutUnban(long userId, MessageEmbed unbanEmbed, ModerationConfig moderationConfig, Message unbanLogMessage) {
+		JDA jda = moderationConfig.getGuild().getJDA();
+		jda.retrieveUserById(userId)
+			.flatMap(User::openPrivateChannel)
+			.flatMap(c -> c.getHistory().retrievePast(1))
```
liquiddevelopmentnet, line 306: if the user sends the bot a dm after the ban message, wouldn't this cause problems? especially later on in the code, with the bot author check

liquiddevelopmentnet (changes requested):

src/main/java/net/discordjug/javabot/systems/moderation/UnbanCommand.java
danthe1st: Fixed in 6922115ee42816b94763db6d920063e928835552.

src/main/java/net/discordjug/javabot/systems/moderation/ModerationService.java
danthe1st: In that case, the update wouldn't be delivered because of the `banMessage.getAuthor().getIdLong() != jda.getSelfUser().getIdLong()` check.
While this could be avoided by retrieving multiple messages (and use the most recent bot messages), users cannot send DMs to the bot when they are banned so the window for that should be really short.

liquiddevelopmentnet: I didn't know the bot doesn't accept dms from non-members 👍

liquiddevelopmentnet (approved):

src/main/java/net/discordjug/javabot/systems/moderation/ModerationService.java
danthe1st: The bot cannot control this but if there is no mutual guild, DMs cannot be sent. This is the same between normal members except that our bot won't be in multiple Discord servers.

